PR #2, opened 2015-04-08
Modified Nmakefile to compile with Visual Studio 2010

dawsonlp:
The current Nmakefile was not compiling for me in Visual Studio 2010, this one compiles in 2010 and 2015 for both 64 bit and 32 bit. At least for the default and the apps target. The 'all' target has some other problem.


freddy77:
Hi,
  Looks good. Why did you remove sources?
I'll merge it when I'll come back from holidays.
Did you try with cmake?

Frediano
Il 08/Apr/2015 22:29, "dawsonlp" <notifications@github.com> ha scritto:

dawsonlp:
I took out the sources because I didn't see any value to them and it made the file easier to read and avoided having to synchronize the OBJ lists with Source lists.  Let me know if that was a bad idea...I can put them back if it makes sense.
Just downloaded and looking at cmake now - I wasn't familiar with it.

dawsonlp:
I've made some other changes to the build that were necessary for me to be able to clone the repository locally to a windows box and in the VS 2010 command prompt simply run the two commands below to build successfully. However, I'm not sure I properly understand the build system that was in place. It looks like config.h should be created automatically, and on linux autotools should take care of that, but how is it supposed to be created on windows?
Commands to build from local git clone:
nmake -f Nmakefile PLATFORM=x64
nmake -f Nmakefile PLATFORM=x64 apps

freddy77:
Hi,
  There is a template win32/config.h.in. Some old projects copied this into
win32/config.h. Would even better to have some script to better generate
the file. Cmake generate a proper config.h.

Frediano
Il 11/Apr/2015 00:41, "dawsonlp" <notifications@github.com> ha scritto:

dawsonlp:
It's taking me a while to get to this, but I'll try and get something
simple in over the next few days based on a simple copy of config.h.in, and
if I can free up some time over the next couple of weeks, I'll look at
putting something less fragile together with cmake.

freddy77:
Cmake is already supported. See userguide on the snapshot  tarball.

Frediano
Il 16/Apr/2015 01:23, "dawsonlp" <notifications@github.com> ha scritto:

ramiro:
Just to make clear the AppVeyor work I mentioned above is so far is on the Branch-0:91 branch.

I started work toward getting master to build in Windows too, always using the Nmakefile. Didn't realize this PR existed (my previous comment was to an isolated commit by @dawsonlp I found via Google but which I couldn't associate to any PR) so there is duplicated work: #4, #5 and #6 are partially similar but apiece and without removal of the source files from the Nmakefile.

#3 and #7 are also small but about cosmetic details.

freddy77:
Applied first 2 changesets.

ramiro:
FYI I've opened #8 with another approach to create win32/config.h (from win32/config.h.in.)

freddy77:
Closed as obsolete, I'm going to remove Nmakefile from master, Who need a Windows compile can use cmake files.

PR #7, opened 2015-05-02
Remove comment about generic Nmakefile rule which wouldn't work anyway.

ramiro:
Also, use correct, consistent 'x64' platform name in help/comments instead of 'win64'.


freddy77:
Should not better to fix Win64 ? Well, I think however that win32 is also a platform while win64 is a set of platforms

freddy77:
I realized now that this patch fix a problem is if not just comment changes. Applied.

PR #8, opened 2015-05-03
Generate win32/config.h too using Perl just like other needed files.

ramiro:


msabramo:
Love to see that you're doing Windows build enhancements, because last time I tried to build FreeTDS on Windows back in 2013, it was a mess requiring a bunch of hacking (see http://marc-abramowitz.com/archives/2013/07/21/compiling-freetds-on-windows/).

The Travis CI build failed for this waiting for a password for some reason:

```
The command "echo -e "
 [global]
 port = ${TDSPORT}
 tds version = ${TDSVER}
 encryption = require" >> /tmp/freetds.conf.local" exited with 0.
$ echo -e "SELECT @@VERSION\nGO\n" | TDSDUMPCONFIG=stdout ./src/apps/tsql -H ${TDSPWDSRV} -p ${TDSPORT} -U $TDSPWDUID -P $TDSPWDPWD | awk '(/encrypt level/ || data)   { print $0 } (/^Microsoft SQL Server/)   { data = 1; print ""; print } (/1 row affected/)          { exit }'
Password: 
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

freddy77:
2015-05-05 2:05 GMT+01:00 Marc Abramowitz <notifications@github.com>:

> Love to see that you're doing Windows build enhancements, because last
> time I tried to build FreeTDS on Windows back in 2013, it was a mess
> requiring a bunch of hacking (see
> http://marc-abramowitz.com/archives/2013/07/21/compiling-freetds-on-windows/
> ).
>
> The Travis CI build failed for this waiting for a password for some reason:
>
> The command "echo -e "
>  [global]
>  port = ${TDSPORT}
>  tds version = ${TDSVER}
>  encryption = require" >> /tmp/freetds.conf.local" exited with 0.
> $ echo -e "SELECT @@VERSION\nGO\n" | TDSDUMPCONFIG=stdout ./src/apps/tsql -H ${TDSPWDSRV} -p ${TDSPORT} -U $TDSPWDUID -P $TDSPWDPWD | awk '(/encrypt level/ || data)   { print $0 } (/^Microsoft SQL Server/)   { data = 1; print ""; print } (/1 row affected/)          { exit }'
> Password:
> No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
> The build has been terminated
>
>
> Quite strange. Could be that the problem is that strings are not quoted
and for some reasons TDSPWDPWD and TDSPWDUID were empty? In this case tsql
would assume that -P is the username and attempt to read password from
standard input. Can you change the command to quote all strings to avoid
this possible issue?

Frediano

freddy77:
Ramiro, I'm doing similar changes to CMake projects file (to compile from git archive and not only from tarball). I would prefer to make CMake the official solution for Windows instead of Nmakefile. On CMake I'm trying to extract the version from configure.ac/configure to avoid having to maintain too different version spread all over the places. Could you do something like what configure does? I noted that Nmakefile is not using version.rc. Possibly is a mistake, I think all dll should use it (or perhaps there is no dll support?).

ramiro:
Frediano, I will try to take a look at building via CMake on Windows starting from a clean git checkout, and to get Nmakefile to use the global version token, hopefully during the next week.

Regarding possibility of building FreeTDS as a set of Windows DLLs, I looked at it a while back.  
From what I've seen, currently it isn't possible. AFAIK the public functions would need to be marked as exportable and some (stub?) entry point function would need to be written. Don't know if the modern MS tools are capable of providing these missing pieces more or less automatically.

freddy77:
Good, I'm not that familiar with Nmake at all. It requires embedding much
more Perl code.

The way is currently done by CMake (and by the VS projects, even if they
are old and I was thinking to remove in a future) is to used the module
definition files (.def). Although they are a quite old way to achieve this
it always been supported by all Windows tools. They are still documented in
Visual Studio 2015!

Frediano

2015-05-07 12:33 GMT+01:00 Ramiro Morales <notifications@github.com>:

ramiro:
@freddy77 I've verified Branch-0_95 and master can be build with CMake (with the "NMake Makefiles" generator) + MS toolchains (Visual Studio 2008 and 2010) for x86/x64

See [1] and [2] for the respective build logs on AppVeyor.

I see building via CMake effectively builds dblib and ctlib as DLLs without any additional effort.

So, my humble proposal would be to simply drop the `Nmakefile`. Rationale for this

* It is unable to build DLL versions of dblib and ctlib
* It's still missing the ability to build some binaries: The bsqlodbc.exe program and the tdsodbc.dll ODBC DLL
* Right now they link in the (MS) C RT libs statically (/MT) which seems a bit weird
* It doesn't build the (many) unittest executables

I've opened #9 to make it possible to also build dblib and ctlib as static libraries with CMake.

---
1. https://ci.appveyor.com/project/ramiro/freetds/build/102/job/dev6342yapx9khiv
2. https://ci.appveyor.com/project/ramiro/freetds/build/101/job/th28glu2itow73gl

freddy77:
I didn't notice all these missing feature and problems with Nmakefile (not that I'm a bit fun of it).
You are probably right, I'll have a look at references (like userguide).
From master I already removed static visual studio projects.

ramiro:
> The Travis CI build failed for this waiting for a password for some reason:

@msabramo I see all PR build fail: https://travis-ci.org/FreeTDS/freetds/pull_requests

Isn't this because we the PR submitters aren't part of the FreeTDS entity at GitHub and so we haven't the needed rights to run the test suite agianst your SQL Server instance (similarly to what happens in pymssql/pymssql)?

freddy77:
Closing as obsolete, Nmakefile will be removed from master  and people can use cmake is they need a build for Windows

PR #9, opened 2015-05-17
Add building of static versions of ct and sybdb on Windows wth CMake.

ramiro:
Names are compatible with historical names used in the deprecated
Nmakefiles and so don't clash with DLL's import libs: libct.lib
and db-lib.lib respectively.

### DIFF
--- a/src/ctlib/CMakeLists.txt
+++ b/src/ctlib/CMakeLists.txt
@@ -1,7 +1,9 @@
 add_subdirectory(unittests)
 
+set(static_lib_name ct)
 if(WIN32)
 	set(win_SRCS ../../win32/initnet.c ctlib.def)
+	set(static_lib_name libct)
 endif()
 
 add_library(ct SHARED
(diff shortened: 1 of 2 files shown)


freddy77:
Is there a reason why you build them only for Windows?

freddy77:
Oh, I noted you included initnet.c and definition files. First should not be included as contains code for DLL (the DLL entry point) while second is ignored by static libraries (well, not an issue).
Only dbopen.c in dblib is necessary but can be avoided. Our dblib is not ABI compatible with ms one.

ramiro:
> Is there a reason why you build them only for Windows?

Actually no. It's just that it occurred to me at the last minute It'd be good to to minimize the impact of the proposed change.

PR updated.

freddy77:
I'll try if new patch work even for Linux
 Il 21/Mag/2015 15:50, "Ramiro Morales" <notifications@github.com> ha
scritto:

freddy77:
Works fine even on Linux. Could you merge the two commits and rebase on last update? Also the name for libct in Unix became something like liblibct.a. Not really nice to see. What about something like ct-lib (similar to db-lib one) ? Or even without the hyphen

ramiro:
> Could you merge the two commits and rebase on last update?

Done.

> Also the name for libct in Unix became something like liblibct.a. Not really nice to see. What about something like ct-lib (similar to db-lib one) ? Or even without the hyphen

I've added some changes  in 44f70a8 for this following a CMake FAQ entry. I've tested it on Linux where it generates a "standard" libct.a (no danger there of crashes with files related to the dynamic one) and on Windows where it still generates the backward-compatible libct.lib.

Also, in ba54c66 a forgotten directive was added.

freddy77:
Merged. I squashed the third in the first, I don't like fixes patch of not committed patch.

Do you think would be sensible to remove Nmakefile ?

I was wondering if is fine to remove old visual studio projects too.

ramiro:
@freddy77 IMHO it'd be sensible to remote Nmakefile in master.

Can I close this PR?

freddy77:
Yes, already merged. I think I'll delete Nmakefile from master and close Nmakefile change requests as obsolete for 0.95

PR #6, opened 2015-05-02
Add packet.c source to tds lib in Nmakefile.

ramiro:
It was missing and caused errors like:
```
db-lib.lib(login.obj) : error LNK2019: unresolved external symbol _tds_read_packet referenced in function _tds71_do_login
db-lib.lib(read.obj) : error LNK2001: unresolved external symbol _tds_read_packet
db-lib.lib(query.obj) : error LNK2019: unresolved external symbol _tds_put_cancel referenced in function _tds_send_cancel
db-lib.lib(net.obj) : error LNK2001: unresolved external symbol _tds_put_cancel
db-lib.lib(write.obj) : error LNK2019: unresolved external symbol _tds_write_packet referenced in function _tds_put_n
db-lib.lib(stream.obj) : error LNK2001: unresolved external symbol _tds_write_packet
```

PR #5, opened 2015-05-02
Add a couple of missing files to replacements helper lib.

ramiro:
They were missing from Nmakefile and caused errors like:

```
bsqldb.obj : error LNK2019: unresolved external symbol _getpassarg referenced in function _get_login
db-lib.lib(mem.obj) : error LNK2019: unresolved external symbol _tds_socketpair referenced in function _tds_init_connection
```

PR #4, opened 2015-05-02
Add dbpivot.c to db-lib sources.

ramiro:
It was missing and this causes errors like:
```
bsqldb.obj : error LNK2019: unresolved external symbol _dbpivot_lookup_name referenced in function _parse_pivot_description
bsqldb.obj : error LNK2019: unresolved external symbol _dbpivot referenced in function _print_results
db-lib.lib(dblib.obj) : error LNK2019: unresolved external symbol _dbnextrow_pivoted referenced in function _dbnextrow
db-lib.lib(dblib.obj) : error LNK2019: unresolved external symbol _dbrows_pivoted referenced in function _dbnextrow
```

PR #3, opened 2015-05-02
Mention ctlib can be built with Nmakefile in help text.

ramiro:
Refs 52e59affc0110c5ddd379ef1e45c4554123f27b5.

PR #10, opened 2015-05-23
CMake: Don't build test programs by default.

ramiro:
Running

  $ [n]make [all]

only builds libs and utiility apps now.

Add new 'check' target which depends on actual building of these
binaries and then runs the test suite via CMake's CTest utility.

See http://www.cmake.org/Wiki/CMakeEmulateMakeCheck


ramiro:
Got this change running on AppVeyor (hosted Windows CI platform).

`nmake` execution (shorter than before because it doesn't build unittest executables):

https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L241

`nmake check` execution:

https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L657

`ctest -V` execution 8called from `nmake check`:

https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L1476

Tests run successfully against SQL Server 2008 r2 sp2 (`"55% tests passed, 69 tests failed out of 155"`), except the ODBC ones. I'm not actually sure it's possible to set up a ODBC testing environment interfacing with the native Windows ODBC stack:

https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L10941

freddy77:
2015-05-24 19:58 GMT+01:00 Ramiro Morales <notifications@github.com>:

> Got this change running on AppVeyor (hosted Windows CI platform).
>
> nmake execution (shorter than before because it doesn't build unittest
> executables):
>
>
> https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L241
>
> nmake check execution:
>
>
> https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L657
>
> ctest -V execution 8called from nmake check:
>
>
> https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L1476
>
> Wonderful!

> Tests run successfully against SQL Server 2008 r2 sp2 ("55% tests passed,
> 69 tests failed out of 155"), except the ODBC ones. I'm not actually sure
> it's possible to set up a ODBC testing environment interfacing with the
> native Windows ODBC stack:
>
>
> https://ci.appveyor.com/project/ramiro/freetds/build/135/job/mjk0a18anxx1wu1q#L10941
>
>
> You should create a proper DSN configuration. You can also register the
odbc driver with regsrv32 command. There should be a odbcconf command you
can use to create a DSN.

Frediano

freddy77:
I don't know if apply this or not. I was expecting some changes to make AppVeyor work correctly.
Are these changes addresses for 0.95 or master?
They actually does not fix any build problem.

ramiro:
I'll open a separate PR for the AppVeyor thing once it's ready. Hopefully before 0.95 goes gold. See branches [appveyor-0.95](https://github.com/ramiro/freetds/tree/appveyor-0.95) and [appveyor](https://github.com/ramiro/freetds/tree/appveyor) for WIP (warning: `git  push --force` often.)

Reasoning sequence for this is:

1. It would be nice to get Windows builds from git checkouts when using MS compilers (because that's would be preferable for pymssql as that's what used for Python officlal Windows binaries). Builds with Mingw seems to be taken care of already.

2. So, PRs to help getting NMakefile in shape were opened (#3, #4, #5, #6, #7, #8).

3. Above PR #8 had not many chances of landing. Cmake would be preferred to build on Windows.OK, submit #9.

4. Although the idea is to take advantage of AppVeyor to generate FreeTDS, it would be great to use it for what's been created: Running tests of the built binaries against a variety of MS SQL Server versions.

5. But using NMakefile or CMake on Windows doesn't provide a way to collect/run the FreeTDS test suite (this is another disadvantage of NMakefile  I forgot to mention in https://github.com/FreeTDS/freetds/pull/8#issuecomment-103041453) as we don't have `make check` facilities provided by autotools. That's why I opted for the easy choice of taking advantage of Cmake's CTest. 

Actually the title of this PR is wrong: It should be "Provide a way to collect/run the test suite on Windows when using MS toolchain". The fact that there is an added "chjeck" make target which builds/runs these tests is a byproduct.

ramiro:
I opened this againt 0.95 But I think it should apply to master as well without conflicts?

One thing I fortogt to mention in item 1 above, is that getting build triggered by Git commits to work without having to perform any additional tweaks is what a CI setup needs but it would be useful for users/contributors too.

freddy77:
If a patch is committed to 0.95 should be ported to master. Actually the right thing would be to develop on master and port to 0.95 but as still rc is fine.
Yes, I understood the story of these patches. Thanks for the effort.

ramiro:
I've opened PR #13 which replaces this one, is opened against master and has a better commit message.

PR #12, opened 2015-06-14
CMakeLists.txt: Detect OpenSSL.

ramiro:


freddy77:
What about this last change? https://github.com/freddy77/freetds/commits/cmake_openssl (is a patch on your patch)
It check for OpenSSL on all platforms.
If you agree and works for you can you squeeze it in yours?

ramiro:
Sure, will test it on Linux and squash commits if it works.

 Didn't dare to make it too general because of the fact CMakeLists.txt
already supported GnuTLS there and didn't know hot to handle the
precedence between both libs. But you've solved that.

On Wed, Jun 17, 2015 at 6:01 PM, freddy77 <notifications@github.com> wrote:
> What about this last change?
> https://github.com/freddy77/freetds/commits/cmake_openssl (is a patch on
> your patch)
> It check for OpenSSL on all platforms.
> If you agree and works for you can you squeeze it in yours?
>
> —
> Reply to this email directly or view it on GitHub.



-- 
Ramiro Morales
@ramiromorales

freddy77:
Well, actually the cmake works better with Windows, would be great to complete for every platform but unfortunately I thing for many Unices autoconf is still the preferred (and perhaps only available) choice.
The definitive way would be that user could be specify OpenSSL or GnuTLS as in autoconf

ramiro:
Ok, I've testes whis on my Ubuntu 14.04 LTS system and it detects and links in OpenSSL correctly (verified with `tsql -C`).

Now, when I uninstall openssl-dev and retry cmake/make I get:

```
Linking C executable tsql
/usr/bin/ld: ../tds/libtds.a(tls.c.o): undefined reference to symbol 'gcry_control@@GCRYPT_1.2'
//lib/x86_64-linux-gnu/libgcrypt.so.11: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [src/apps/tsql] Error 1
make[1]: *** [src/apps/CMakeFiles/tsql.dir/all] Error 2
make: *** [all] Error 2
```
From e.g. https://bugzilla.redhat.com/show_bug.cgi?id=625459#c1 seems to be the fact that `gcrypt` is missing [here](https://github.com/freddy77/freetds/commit/24cc41ef673e3ac600461919451678669aa5dec7#diff-af3b638bc2a3e6c650974192a53c7291R284)? If I add it back then the build is successful and `tsql -C` reports GnuTLS is in use.

freddy77:
Yes, I removed gcrypt reference but was a mistake. Added again.
I'm using a quite new version of GnuTLS which use nettle instead of gcrypt and does not require gcry_control call

ramiro:
I've updated, squashed and rebased this on your recent work on on master.

PR #13, opened 2015-06-14
Collect, run test on Windows+CMake

ramiro:
Now, running

  $ [n]make [all]

only builds FreeTDS libs and utility apps.

Add a new 'check' target which depends on actual building of these
binaries and then runs the test suite via CMake's CTest utility.

See http://www.cmake.org/Wiki/CMakeEmulateMakeCheck


freddy77:
Applied!

PR #16, opened 2015-07-02
Branch 0 95

rzsis:
I get compile dblib_dll 0.91rc in c++2010.
I try compile dblib_dll.dll 0.95 in c++ 2010,but in 0.95 version don't have a msvc6 folder, I update 0.91rc project to 0.95 folder , but when I compile dblib_dll I get the follow error and many other LNK2019 error
2>libTDS.lib(data.obj) : error LNK2019: unresolved external _tds_copy_stream referenced in function _tds_get_char_dynamic.symbol 

Tds is not more compatible with microsoft compiler?
If possible I would like help to contribuite compatible to make dblib_dll with c++, but i will make some help to do it


freddy77:
Yes, we are moving to cmake, see userguide

PR #17, opened 2015-07-13
AppVeyor: Auto-detect current version of OpenSSL binaries.

ramiro:
As OpenSSL is a rapidly moving target, this allows us to avoid having to
manually and periodically set the current version in the appveyor.yml
file.

Implement this by processing the Shining Light Productions downloads
page with a command-line utility that allows us to apply XPath queries
on it (xidel: http://www.videlibri.de/xidel.html) to extract the version
number/token.


freddy77:
Merged

PR #19, opened 2015-07-31
[don't merge] Appveyor: Visual Studio 2015

ramiro:
Just to trigger a visible build showing CMake  vsnprint/_vsnprintf detection results with this toolchain (AFAICT still no dice.)


freddy77:
Merged with small fixes.

PR #21, opened 2015-09-07
This change improves compatibility with a great number of SQL Server …

kozmaz87:
…versions. No encrypted connection works without either gnutls or openssl. In Microsoft Azure encryption is required so users trying to use the generated RPM package find themselves getting the error message: Adaptive server connection failed. With tds version = auto and some crypto library support the issue can be solved. Since openssl is not compatible with the project licence I understand that --with-openssl is not an option for default but --with-gnutls might be a possible default option.

Questions: why cannot one configure --with-tdsver=auto? What is wrong with the auto trying algorithm? It seems really discouraged by the looks of it...

### DIFF
--- a/freetds.conf
+++ b/freetds.conf
@@ -10,7 +10,7 @@
 # server specific section
 [global]
         # TDS protocol version
-;	tds version = 4.2
+	tds version = auto
 
 	# Whether to write a TDSDUMP file for diagnostic purposes
 	# (setting this to /tmp is insecure on a multi-user system)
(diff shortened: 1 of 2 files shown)


freddy77:
I like your patch, probably it will help reducing the mail for improper protocol.
Although not optimal for connection it works.
I added some additional patches, see last 4 commits in https://github.com/freddy77/freetds/commits/rpm_compatiblity
Beside the first patch I think the other could be changes to your patches.
- there is a bug in your call to configure in the spec file
- you should add dependencies to the spec to enable mock or other tool to compile the package
- you didn't add any changelog to the spec file
- you should split your commit message (in git) to subject + detailed message.

freddy77:
No reply in about 2 months, I'll integrate the changes as I like

freddy77:
Fixed, improved and merged.

PR #22, opened 2015-10-05
Appveyor: Change 3rd. party OpenSSL Windows binaries used.

ramiro:
Switch to use the ones published at
http://www.npcglib.org/~stathis/blog/precompiled-openssl/

The ones from slproweb.com are built with VS2013 now but we are building
FreeTDS with VS 2008, 2010 and 2015.

So we have the following dependency chain:

- FreeTDS (both as static libs and dynamic libs)
- OpenSSL (as dynamic libs, because that's how we are building FreeTDS)
- The Visual C runtime libs (as dynamic libs, because that's how we are
  building FreeTDS and OpenSSL)

When the C runtime libs are linked in dynamically, a mismatch of
versions is a recipe for disaster. So we need to make sure FreeTDS and
OpenSSL are built using the same MS toolchain version.

### DIFF
--- a/misc/appveyor.yml
+++ b/misc/appveyor.yml
@@ -27,6 +27,7 @@ environment:
     # See: http://stackoverflow.com/a/13751649/163740
     WITH_COMPILER: "cmd /E:ON /V:ON /C ..\\misc\\run_with_compiler.cmd"
     WIN_ICONV_VER: 0.0.6
+    OPENSSL_VER: 1.0.2d
 
   matrix:
   - PLAT: Win32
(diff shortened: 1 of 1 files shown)


freddy77:
The problem with this website is:
- there is no SSL version so it's not 100% sure you are downloading from the real website;
- this patch remove support for detecting last OpenSSL version so if we forget to check and update you can get insecure executables (well, ABI should be kept but it's not sure).

About the recipe for disaster consider that a lot of system dlls have this issue so if this would be really a problem probably Windows application would crash quite often. If you don't mix with different structures or allocation/deallocation function is not a problem. OpenSSL has been around for quite a while as Windows library so I don't thing they have these issues.

We don't provide OpenSSL so the user, if statically linking is required, has to download the proper OpenSSL library which are supposed to be binary compatible (C is one of the better language for ABI compatibility).

ramiro:
> The problem with this website is:
> 
>     there is no SSL version so it's not 100% sure you are downloading from the real website;

Right. That's a problem.

>     this patch remove support for detecting last OpenSSL version so if we forget to check and update you can get insecure executables (well, ABI should be kept but it's not sure).

Yes, had to undo that trick, unfortunately.

> 
> About the recipe for disaster consider that a lot of system dlls have this issue so if this would be really a problem probably Windows application would crash quite often. If you don't mix with different structures or allocation/deallocation function is not a problem. OpenSSL has been around for quite a while as Windows library so I don't thing they have these issues.

I must confess I was surprised no such cross-C runtime memory operations problems had happened so far. Was just trying to follow common advice when it comes to MSVC runtime versions as I used to get bitten by such issues a few years ago at work, but has no real world evidence it was causing trouble for our CI setup. 

I actually have relatively limted experience with both recent MS C toolchains and OpenSSL on Windows so I'm ok with not touching anything on this front.

Please feel free to close this PR.

freddy77:
Closing the PR as agreed.

PR #23, opened 2015-10-13
Appveyor CI setup for 0.95 branch

ramiro:
I plan to resume work on getting pymssql WIndows binaries automatically built via AppVeyor.

For this I'll need FreeTDS most recent stable release built with the three MS C compilers used by supported official Windows Python versions (VS 2008, 2010 and 2015).

Also, as pymssql announces Azure connectivity compatibility, the FreeTDS libs used needs to be linked against OpenSSL.

All of this we already have on master, but I'll need it for 0.95.

This PR contains a baseline composed of 42 commits cherry-picked from master and applied in the same chronological order over `Branch-0_95`. These commits are the ones needed to get the 0.95 test suite to run on AppVeyor with the features described above plus 'feature parity' with `master`.

Only four non-unittest, non-AppVeyor, non-CMake files are touched wth minor changes.

Maybe merging this isn't a good idea, maybe it's better to keep it as a separate branch so it's easily rebasable againt branch `Branch-0_95`.

Or maybe this isn't worth the trouble and this PR can be closed right away. Please feel free to do so it you think it's better.

I figured it could be a good idea to open this PR and submit it to your consideration at this point because, for the pymssqI work, I plan to hack more heavily and get things to actually diverge by committing changes that don't originate on master (will do this work on another branch without an associated PR.)


ramiro:
Forgot to add a link to AppVeyor CI results. here it is: https://ci.appveyor.com/project/ramiro/freetds/build/324

freddy77:
Applied!
Usually I don't accept any no-fix change but is useful to have and no-test changes are really limited.
A bit worried about next huge changes, I hope they will be in python side, not in FreeTDS one.

PR #24, opened 2015-10-15
Missing header files

o-rey:
Seems like couple of headers are missing, so the code does not compile.


freddy77:
Applied (remove space changes)

PR #25, opened 2015-10-18
Remove unused includes for file wspiapi.h

larskanis:
This popped up with an older Mingw version, which didn't include this header file. However there is no `Wspiapi*` function used anywhere in the FreeTDS code. So this include could be removed.


freddy77:
Merged, thanks.

PR #26, opened 2015-10-18
Backport 'Avoid problem with getpid definition'

larskanis:
Include `unistd.h` before `replacements.h` to avoid problems with the macro we define for `getpid`.

This is already fixed on the master branch: https://github.com/FreeTDS/freetds/commit/c5cd0682577


freddy77:
Hi, applied directly original patch

thanks.

PR #31, opened 2015-11-05
Implemented the "ApplicationIntent" property that is was introduced in TDS 7.4

tapas-sharma:
Implemented the "ApplicationIntent" property that is new in TDS 7.4 in the current stable version of "freetds_0.91".

You can use this by putting "ApplicationIntent=ReadOnly", in the connection string.
If you do not set this property the default value is ReadWrite, in which case the user can fire write queries too.
Remember to set the following as connection string if you want to do any WinAuth connections

"DRIVER=FreeTDS;Server=<IP>;Port=<PORT>;Database=<DB_NAME>;UID=<USERNAME>;PWD=<PASSWORD>;ApplicationIntent=<ReadWrite/ReadOnly>;TDS_Version=7.2;"

TODO:
Add test in the php/python.
Try to merge with the master branch and the current codeline.

### DIFF
--- a/README
+++ b/README
@@ -36,7 +36,13 @@ A User Guide, in sgml and html form, is included in this distribution.
 Also included is a reference manual, generated in HTML with Doxygen. 
 "make install" installs the HTML documentation, by default to
 /usr/local/share/doc/freetds-<version>.  
+ 
+Implemented the "ApplicationIntent" property that is new in TDS 7.4 in the current stable version of "freetds_0.91".
+You can use this by putting "ApplicationIntent=ReadOnly", in the connection string.
+If you do not set this property the default value is ReadWrite, in which case the user can fire write queries too.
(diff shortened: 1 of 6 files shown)


freddy77:
Please note also that last version is 0.95 not 0.91.
This patch is adding a new feature so you should address master branch.

If you want I can propose an updated patch

tapas-sharma:
@freddy77 Surely.

freddy77:
What about this? https://github.com/freddy77/freetds/commit/639f072707020b48b3c6423fbdadd3a9b63a7c5d

freddy77:
Updated to this https://github.com/freddy77/freetds/commit/fe93a8a53150068696906e0eab9f0d7ff35543f5.
Changes:
- added documentation in userguide;
- fails to connect if ApplicationIntent (ODBC) is not one of the expected values.

freddy77:
Merged

PR #34, opened 2015-12-02
Store the SPID value returned from the last seen TDS packet.

joshuahlang:
The SPID value wasn't working with the db-lib api dbspid() method. This value can be useful for debugging, so I add an implementation for it. The SPID value is pulled from the last seen TDS packet header. See https://msdn.microsoft.com/en-us/library/dd305266.aspx.

I don't have much experience in the code, so perhaps there is a more appropriate way to get this value?


freddy77:
Merged!

Thanks!

PR #38, opened 2015-12-14
Fixed the LOGIN and Connection issue with TDSVER >= 7.2

johnnyzhong:
Hi, I just checked out the code from master branch, and compare the change, 

I think we have some common fixing in different code, and I suggest to merge our code, because:

1. I also fixed some minor bugs in the TOKEN length value, which seems not impact the finally result, but anyways it is not correct, and it will impact debugging in future, 
because the correct length field will help us clearly analyze the network packet decoded as TDS protocol in Wireshark.

2. For the change in tds7_read_login(), 

a) we both fixed the login issues for ver >= 7.2, but  I saw that they wrote a constant digital number there as the offset to get the string of Host Name, User Name, Pasword, etc.

currently it is correct for 7.1 - 7.4, 
but actually the offset value is stored in a specific byte in the header, we'd better get the correct offset from that Byte instead of a constant value, in case of upgrading protocol, the offset will change just like from 7.1 to 7.2.

b) I also fixed a minor issue to retrieve the correct TDS_VERSION from the login packet, there was byte order issue, that is exactly why before you saw the 'ver: **' value printed in console is not correct when testing.

### DIFF
--- a/src/server/login.c
+++ b/src/server/login.c
@@ -167,16 +167,29 @@ tds7_read_login(TDSSOCKET * tds, TDSLOGIN * login)
 	char *pbuf;
 	DSTR database = DSTR_INITIALIZER;
 	int res = 1;
+    /* BEGIN: Added by Johnny Zhong, 2015/12/13 */
+    int host_name_offset = 0;
+    /* END:   Added by Johnny Zhong, 2015/12/13 */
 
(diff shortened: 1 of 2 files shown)


freddy77:
Please update on last mster.

PR #39, opened 2015-12-14
Fixed the LOGIN and Connection issue with TDSVER >= 7.2, and correct the length field of specific TOKENs

johnnyzhong:
I think we have some common fixing in different code, and I suggest to merge our code, because:

1. I also fixed some minor bugs in the TOKEN length value, which seems not impact the finally result, but anyways it is not correct, and it will impact debugging in future, 
because the correct length field will help us clearly analyze the network packet decoded as TDS protocol in Wireshark.

2. For the change in tds7_read_login(), 

a) we both fixed the login issues for ver >= 7.2, but  I saw that they wrote a constant digital number there as the offset to get the string of Host Name, User Name, Pasword, etc.

currently it is correct for 7.1 - 7.4, 
but actually the offset value is stored in a specific byte in the header, we'd better get the correct offset from that Byte instead of a constant value, in case of upgrading protocol, the offset will change just like from 7.1 to 7.2.

b) I also fixed a minor issue that, retrieve the correct TDS_VERSION from the login packet, there was byte order issue, that is exactly why before you saw the 'ver: **' value printed in console is not correct when testing.

### DIFF
--- a/src/server/login.c
+++ b/src/server/login.c
@@ -167,16 +167,29 @@ tds7_read_login(TDSSOCKET * tds, TDSLOGIN * login)
 	char *pbuf;
 	DSTR database = DSTR_INITIALIZER;
 	int res = 1;
+    /* BEGIN: Added by Johnny Zhong, 2015/12/13 */
+    int host_name_offset = 0;
+    /* END:   Added by Johnny Zhong, 2015/12/13 */
 
 	a = tds_get_int(tds);	/*total packet size */
 	a = tds_get_int(tds);	/*TDS version */
+    
+    /* BEGIN: Added by Johnny Zhong, 2015/12/12 */
+    a = ntohl(a);
+    /* END:   Added by Johnny Zhong, 2015/12/12 */
+    
 	if ((a & 0xff) == 7)
 		tds_set_version(login, a & 0xff, (a >> 8) & 0xff);
 	else
 		tds_set_version(login, (a >> 4) & 0xf, a & 0xf);
 	a = tds_get_int(tds);	/*desired packet size being requested by client */
 	tds_get_n(tds, NULL, 24);	/*magic1 */
-	a = tds_get_smallint(tds);	/*current position */
+    
+    /* BEGIN: Added by Johnny Zhong, 2015/12/13 */
+    /* client name offset from the start of LOGIN PACKET (the concept is different from the TCP packet) */
+	host_name_offset = tds_get_smallint(tds);	
+    /* END:   Added by Johnny Zhong, 2015/12/13 */
+    
 	host_name_len = tds_get_smallint(tds);
 	a = tds_get_smallint(tds);	/*current position */
 	user_name_len = tds_get_smallint(tds);
@@ -200,6 +213,11 @@ tds7_read_login(TDSSOCKET * tds, TDSLOGIN * login)
 	a = tds_get_smallint(tds);	/*total packet size */
 	tds_get_smallint(tds);
 
+    /* BEGIN: Added by Johnny Zhong, 2015/12/11 */
+    /* There are 8 bytes before the LOGIN packet (type, status, length, ..., etc) */ 
+    tds->in_pos = host_name_offset + 8;
+    /* END:   Added by Johnny Zhong, 2015/12/11 */
+    
 	res = res && tds_dstr_get(tds, &login->client_host_name, host_name_len);
 	res = res && tds_dstr_get(tds, &login->user_name, user_name_len);
 
--- a/src/server/login.c
+++ b/src/server/login.c
@@ -159,64 +159,91 @@ int
 tds7_read_login(TDSSOCKET * tds, TDSLOGIN * login)
 {
 	int a;
-	int host_name_len, user_name_len, app_name_len, server_name_len;
-	int library_name_len, language_name_len;
-	int auth_len, database_name_len;
+	unsigned host_name_len, user_name_len, app_name_len, server_name_len;
+	unsigned library_name_len, language_name_len;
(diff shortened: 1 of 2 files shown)


freddy77:
These patches are still base on an old master version.

Note on style:
- use tabs;
- brackets follow condition at same line;
- do not annotate all changes, we use git to track changes.

freddy77:
I think these are the diff left https://github.com/freddy77/freetds/commit/a65d7f030d7be6f9d2dac1a52cc9943a507e6b47

The change for tds_send_msg are perhaps a bit too long, just changing 
  + (IS_TDS7_PLUS(tds->conn) ? 2 : 1) * (strlen(msgtext) + 1 + strlen(srvname) + 1 + len)
to 
  + 4 + (IS_TDS7_PLUS(tds->conn) ? 2 : 1) * (strlen(msgtext) + strlen(srvname) + len)
Should support all versions. Note that you added 1 for token type but is correct only as the message length is 2 and not 1.

The change for tds_env_change looks good beside you should check for protocol version (from 7.0).
I don't think this function is correct for second change (well, wasn't correct even before anyway).

For tds7_read_login:
- ntohl should not be used as it create problems with big endian machines, emul_little_endian is enabled so you should get same number for little or big endian;
- your code remove any checks for host;
- setting in_pos is scaring, you could be over the packet. I readed documentation and there is no specification for offsets validation, only they point to data after the fix part. I think would be better (and easier now) to allocated a memory buffer, read all data after fixed part and use this buffer.

johnnyzhong:
1.  For change to tds_send_msg, I checked it with a TDS sepc from Microsoft and carefully compared with the packets decoded as TDS in Wireshark.

johnnyzhong:
2. About the change for tds_env_change, I also checked the MS TDS spec and check the packets captured in Wireshark, after I change the length, Wireshark can perfectly decode the packets as TDS and each TOKEN are shown very clear and looks correct.

johnnyzhong:
For setting in_pos, to be honest, I am not very clear about how the socket related functions process parse the packets and, but I am sure according to the MS TDS spec, my code logic is correct to calculate the offset.

as for 'it may be over the packet', for the current code, yes here we need to add a boundary check.

Just for your reference, generally using constant value (magic number) is not a good idea I think as long as we can avoid to use Magic Number,first it is not readable, second it may always need to maintained when the protocol upgraded that the offset changed.

johnnyzhong:
For "The change for tds_send_msg are perhaps a bit too long, just changing 
+ (IS_TDS7_PLUS(tds->conn) ? 2 : 1) * (strlen(msgtext) + 1 + strlen(srvname) + 1 + len)
to 
+ 4 + (IS_TDS7_PLUS(tds->conn) ? 2 : 1) * (strlen(msgtext) + strlen(srvname) + len)
Should support all versions. Note that you added 1 for token type but is correct only as the message length is 2 and not 1.", 

I don't understand what is the 'only as the message length is 2 and not 1" referring to?

and I tested my code with 7.1 through 7.4, it is working for 7.1 - 7.4, they worked well, 
and I captured the network packet with tcpdump for 7.1 / 7.2 / 7.3 /7.4, then profiling them in Wireshark and decode as TDS, the result looks pretty good,

freddy77:
The code is supposed to work from TDS 4.2 to 7.4 (well... I'd like to anyway).

The "only as the message length is 2 and not 1" refers to the fact that in your compute you added 1 byte for "Token type" and 1 byte for "Error message length". But token type should not be accounted while message length is a US_VARCHAR on documentation so the length is 2 bytes. The results is the same (and also the assembly code), just the source is not correct.

About in_pos. in_pos is the offset into the network packet. However the login packet can be much bigger, split into multiple network packets. So for instance you could have host_name_offset pointing to 60k on a 4k buffer. Usually (and the current code assume it) buffers are contiguous starting after fixed data.

Which "magic number" are you referring to? The fixed size lengths (like 86)?

Thanks for testing.

johnnyzhong:
freddy, 

I think my code added in tds_send_msg is only for TDSVER >= 7.2, and supposing the existing code is correct for < 7.2, I still don't have a chance to work with VER > 7.1, so I just tried not to ruin the existing logic, and I understand your suggestion is right.

And yes I mean the lengths like 86, it is really not a problem for now.
and I do know generally application messages (packets) can be much bigger and split into multiple network packets, it depends on how the program handles the situation, if my strategy her using in_pos is nor proper for big login packet here, just ignore it.

I am very glad to have a chance to contribute to freetds open source project and work with you, I am actually totally new to TDS and freetds, I just started study and research on freetds several days ago, so I think I need some time to have a overall comprehension to all the details of the code, as well as some specific rules, code style, etc.

For any thing I don't know or any convention of this project, please point out to me, I will be very happy to adjust myself, and pay attention to them. 

Thanks

johnnyzhong:
freddy, 

I think my code added in tds_send_msg is only for TDSVER >= 7.2, and supposing the existing code is correct for < 7.2, I still don't have a chance to work with VER > 7.1, so I just tried not to ruin the existing logic, and I understand your suggestion is right.

And yes I mean the lengths like 86, it is really not a problem for now.
and I do know generally application messages (packets) can be much bigger and split into multiple network packets, it depends on how the program handles the situation, if my strategy her using in_pos is nor proper for big login packet here, just ignore it.

I am very glad to have a chance to contribute to freetds open source project and work with you, I am actually totally new to TDS and freetds, I just started study and research on freetds several days ago, so I think I need some time to have a overall comprehension to all the details of the code, as well as some specific rules, code style, etc.

For any thing I don't know or any convention of this project, please point out to me, I will be very happy to adjust myself, and pay attention to them. 

Thanks

PR #48, opened 2016-01-28
Increased max database name length to 128

hanky6:
If a ‘USE <database name larger than 30 characters’ statement is executed,
the dbname function would be trimmed. 128 is based on the microsoft
create database docs.


freddy77:
Looks sensible. The macro is only used for internal static buffer allocations.
I'm not sure about changing or not public interface on 0.95.
Also 128 can be too few if we want to support all charsets

freddy77:
Merged. I moved macro definition to dblib.h (private header)

PR #53, opened 2016-02-28
CMake: Add WITH_OPENSSL option to ignore found OpenSSL.

ramiro:


freddy77:
No comment! Patch is perfect, merged.
I think option could be used instead of others set in the CMakeLists.txt file, but this is another todo...

PR #59, opened 2016-04-05
Add dbgettimeout and dbsettimeout APIs

joshuahlang:
These APIs read and write a tds socket query timeout and can be used to configure a per-connection query timeout instead of using the global timeout value that was set when the connection was created.

### DIFF
--- a/include/sybdb.h
+++ b/include/sybdb.h
@@ -826,6 +826,9 @@ int dbnumrets(DBPROCESS * dbproc);
 DBPROCESS *tdsdbopen(LOGINREC * login, const char *server, int msdblib);
 DBPROCESS *dbopen(LOGINREC * login, const char *server);
 
+RETCODE dbsettimeout(DBPROCESS * dbproc, int timeout);
+int dbgettimeout(DBPROCESS * dbproc);
+
 /* pivot functions */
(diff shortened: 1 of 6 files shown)


freddy77:
Was thinking if was not better to use dbsetllong so to not add any APIs

joshuahlang:
That seems reasonable to me as well. I chose the new APIs for consistency with the existing APIs for setting the global timeouts.

ramiro:
Oops, I hadn't seen this PR and opened #64 today to propose something similar but using the `dbsetopt()` function just like (the very lightly documented) Microsoft SQL Server 2000 DB-Library: https://technet.microsoft.com/en-us/library/aa937147%28v=sql.80%29.aspx?f=255&MSPPError=-2147217396

It seems Sybase doesn't have such an option.

Got driven through that path by a thread on the mailing list from 2009 (http://marc.info/?l=freetds&m=126037476326260) I found by researching about the topic.

We (pymssql) have a feature request open (pymssql/pymssql#410) for which this would be the basis for a fix.

BTW, The OP on that 2009 thread is a co-maintainer of pymssql.

freddy77:
Merged Ramiros patches

PR #61, opened 2016-04-18
Fix mingw-w64 compatibility

larskanis:
Older Mingw-w64 versions redefine `optreset` to `__mingw_optreset`, althought the variable isn't defined or used. The test therefore wrongly succeeds, when `optreset` is declared after `#include <getopt.h>`.

This should fix https://github.com/rails-sqlserver/tiny_tds/pull/265#issuecomment-211329480 .

### DIFF
--- a/include/replacements.h
+++ b/include/replacements.h
@@ -139,6 +139,14 @@ char *tds_basename(char *path);
 int tds_gettimeofday (struct timeval *tv, void *tz);
 #define gettimeofday tds_gettimeofday
 
+/* Older Mingw-w64 versions don't define these flags. */
+#if defined(__MINGW32__) && !defined(AI_ADDRCONFIG)
+#  define AI_ADDRCONFIG 0x00000400
+#endif
+#if defined(__MINGW32__) && !defined(AI_V4MAPPED)
+#  define AI_V4MAPPED 0x00000800
+#endif
+
 #endif
 
 #if !HAVE_GETOPT
--- a/configure.ac
+++ b/configure.ac
@@ -424,7 +424,7 @@ TYPE_SOCKLEN_T
 AC_MSG_CHECKING([whether getopt has optreset support])
 AC_TRY_LINK([#ifdef HAVE_GETOPT_H
 #include <getopt.h>
-#endif], [extern int optreset; optreset = 0;],
+#endif], [optreset = 0;],
 [AC_MSG_RESULT(yes)
 AC_DEFINE(HAVE_GETOPT_OPTRESET, 1, [Define if your getopt(3) defines and uses optreset])],
 [AC_MSG_RESULT(no)]


freddy77:
Why optreset is not working with former MingW versions? How is defined? Could be that we can make it work even if redefined?

The "this should fix" comment is scaring me. Is it working or not?

include/replacements.h
```diff
@@ -139,6 +139,14 @@ char *tds_basename(char *path);
 int tds_gettimeofday (struct timeval *tv, void *tz);
 #define gettimeofday tds_gettimeofday
 
+/* Older Mingw-w64 versions don't define these flags. */
```
freddy77: I think an additional check to check for MingW would be a good thing.

larskanis:
> Why optreset is not working with former MingW versions? How is defined? Could be that we can make it work even if redefined?

optreset isn't defined in mingw unless `_BSD_SOURCE` is set in front of `#include <getopt.h>`. And we don't set `_BSD_SOURCE` anywhere, currently. However before [this mingw-commit](https://github.com/mirror/mingw-w64/commit/603c9f034f4dcfc7c0beeb33cae218620693a69f) , all occurrences of `optreset` were redefined to `__mingw_optreset`, regardless of `_BSD_SOURCE`. This leads to the wrongly success of the test.

After a second though I removed `external int optreset;` entirely, since this should be defined in getopt.h already, and is another option to fix this mingw issue.

> The "this should fix" comment is scaring me. Is it working or not?

Truly, I also wouldn't merge a PR with such a comment... Yes, it's working. Freetds builds fine (with this PR) in the RubyInstaller cross build environment and tiny_tds's `rake gem:windows` runs fine, if I also patch `configure` in addition to `configure.ac` (which isn't in git, but in the tarball). I only wasn't fully sure, whether the PR fixes all the issues that @metaskills refers to.

In addition I tested the build of freetds/tiny_tds with `CFLAGS=-D_BSD_SOURCE`. This successfully enables the optreset feature in freetds, and builds fine as well.

include/replacements.h
larskanis: Mingw-check is added, now.

freddy77:
I pushed the getaddrinfo patch.
Still not clear why the test on optreset does not work. The test, as tsql include getopt.h so should see the same flag. MingW for compability always exports __mingw_optreset. The test should so detect the variable and link to it. With updated MingW variable should not be detected and not used.
So.. what's the issue? Building failure? tsql not working?

metaskills:
Thans for working this issue. I am relying on Lars to communicate the needs or answer questions. All I know is that our CI has been able to build FreeTDS for the Ruby gem till the 1.0rc1.

freddy77:
Ok, finally got a reproduction. Added some notes to the commit.

larskanis:
@freddy77 The optreset issue wasn't easy to describe, but you did it well in your commit message. Thanks for looking at it and merging!

metaskills:
Just wanted to report back that rc3 with these did the trick!

PR #67, opened 2016-05-19
Added dbacolname

hankinsoft:
Added dbacolname so that column names can be retrieved for computed
columns.

### DIFF
--- a/include/freetds/tds.h
+++ b/include/freetds/tds.h
@@ -587,6 +587,7 @@ typedef struct tds_login
 	unsigned int emul_little_endian:1;
 	unsigned int gssapi_use_delegation:1;
 	unsigned int use_ntlmv2:1;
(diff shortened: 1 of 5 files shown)


freddy77:
Patch looks good. I would change the API to return a const char * instead.
I'll add to master.

hankinsoft:
I modified the result to be const char * as per your recommendation.

freddy77:
Thanks. Merged which minor changes (mostly spaces).

hankinsoft:
Thanks! Just to note, I think I added a second change regarding NTLMv2 to this that I should have probably split out. I didn't realize it would be added to my same pull request (not a bit github user).

freddy77:
Merged last patch.
Thanks, sorry for the delay.

PR #71, opened 2016-07-08
Update documentation to provide table for properly choosing a TDS Ver…

FlipperPA:
…sion to use with SQL Server.

### DIFF
--- a/doc/userguide.sgml
+++ b/doc/userguide.sgml
@@ -787,6 +787,88 @@ or
 					</tgroup>
 				</table></para>
 
+<para>
+	Choosing the correct <acronym>TDS</> protocol version for use with SQL Server can be confusing. Hopefully, these steps will lead you to the correct version.
+</para>
+
+<para>
+	<ItemizedList>
+		<listitem>Step 1: Find out which FreeTDS version you are running. You can use the command: `tsql -C`</listitem>
+		<listitem>Step 2: Find out what version of SQL Server you are running.</listitem>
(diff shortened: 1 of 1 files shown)


freddy77:
I don't know why compilation failed... I'll try.
Also I noted there is no documentation for Sybase servers.

FlipperPA:
Thanks, I was wondering about the failues as well, since it was just simple documentation. Apologies for the lack of Sybase, but I have never worked with Sybase, just MS SQL Server. I answer a lot of questions on this topic on Stack Overflow and noticed there was a ton of confusion around TDS versions and SQL Server (most of the questions involved it in one form or another) so I decided a PR might help. If you take a look at a few of my answers, you can see the pattern: http://stackoverflow.com/users/1394697/flipperpa?tab=answers

Thanks for all of your amazing work on FreeTDS.

freddy77:
I think would be enough to change "SQL Server" to "Microsoft SQL Server" and put a line like

"If you have a Sybase server use version 5.0 otherwide if you have Microsoft SQL Server you can follow following section."

well.. or something better than this.

FlipperPA:
I've updated the text to include the sentence about Sybase. I hope this helps!

freddy77:
Merged. I removed the "Microsoft Microsoft SQL Server" and the "Microsoft" where "SQL Server" was referring to any SQL Server (Sybase or Microsoft).
I also split some too long line so on next diff they will be smaller.

PR #83, opened 2016-11-14
Fisql performance issues

Toghr:
Hello,

I noticed that fisql is loosing quite a lot of time in padding spaces to varchar results. We could improve this by simply replacing for loop with memset. In my tests I got significant improvements in performance.
Tests were done on query that selects > 100 columns, 100K rows. Execution times in seconds:
for ~ 34s
memset - 11s

On my setup this difference was even bigger You compile freetds with -m64.

Feel free to drop me an email if You need more details.

Best Regards,
Wojciech Sztachanski


freddy77:
Merged. Minor space changes.

PR #84, opened 2016-11-17
Batch statements without select are not processed correctly

moogah:
If you send a batch of statements in a single call to the database, and these statements don't return rows, the dbresults function doesn't process the data correctly.  It will loop over all data the first time it is called and return to the calling function that there are no more rows.


freddy77:
I'll try but as far as I remember it's the dblib behavior.

freddy77:
Looks like patch is quite working.
You missed entirely to add the tests to Makefiles and so on so are not tested automatically.

I wrote this commit https://github.com/FreeTDS/freetds/commit/f21cd16e8fc19212b79f0cdb89bedb78731ed541 to update unit tests indentation, some minor issues (like variable declaration inside code) and added makefiles.

What's the wf_ prefix ?

The comments are really too short... you should put some more details, for instance reference to the issue

moogah:
I can't say with 100% certainty that having pdo_dblib use a different set of methods wouldn't solve the issue, but while testing I wasn't able to find another set of methods that produced better results.

moogah:
Thanks for taking a look, and yes your feedback is exactly right.  I built this commit attempting to only send in the important files after a couple weeks of testing.. but I removed too much.  I will update my comments and clean up a few things then push to this PR.  Thanks!

moogah:
What are the indentation rules you're using?  I need to tweak the settings in my editor so it stops overwriting them when saving.

freddy77:
Mostly using tab with 8 columns.
In src directory there is an .indent.pro file (the configuration for indent program), you can try to use.
I tested your unit tests with Sybase dblib and got same results. I'm getting some issues with 2 tests I have, still to understand if it's a problem of the tests or not.
I had to change a bit the .sql syntax to support earlier sql servers and sybase.

freddy77:
Could you rebase on master? Development should be done on master.
Can you replace the fprintf(stdout, with just a printf( ?
The $Id stuff at the beginning are old CVS stuff, they are still there in files mostly unchanged.
Also don't use C++ comments (// xxx), only C (/* xxx */).

Unit tests are really fine. The fix introduce some issues on other tests, I'm trying to find a solution to fix them all, seems not so easy.

moogah:
Would it be better if I created a new PR against the master branch as well?  Thanks for your help with this!

-Jeff

freddy77:
There must be some mistake, you merged all changes from master to Branch-1_00.
Tried to remove all patches from master and base on master.
You can find the result at https://github.com/freddy77/freetds/tree/merge3. I would skip the last change (the fix) and squash all other unit tests related together. I would put a comment like:

"Add unit tests for batch statements

Signed-off-by: Jeff Farr <jefarr@wayfair.com>"

moogah:
Hmm, I apologize for the confusion, I intended to rebase to master and I'm not sure exactly where I made the mistake.  I've created another branch based on master and merged in all of my and your changes, except the final fix to dblib.c (did you not want that merged into master yet?).  

Again, sorry for the messy git work, but the next PR is a clean version without any confusing merges in it's history.

https://github.com/FreeTDS/freetds/pull/87

PR #87, opened 2016-11-28
Add unit tests for batch statements

moogah:

Signed-off-by: Jeff Farr jefarr@wayfair.com

### DIFF
--- a/src/dblib/unittests/.gitignore
+++ b/src/dblib/unittests/.gitignore
@@ -43,3 +43,5 @@ pending
 cancel
 spid
 canquery
+batch_stmt_ins_sel
+batch_stmt_ins_upd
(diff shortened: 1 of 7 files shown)


freddy77:
Merged! Now back at the fix. Your patch works for the new test but break rpc test using Sybase.
The bad thing is that using sybase libraries all 3 tests passes and the main issue is that your patch for rpc caused rpc test to return an extra empty recordset with same columns at the previous.

moogah:
I don't have access to a Sybase server to test against and I'm not sure right now how I could go about getting an installation.  The RPC test was also the most difficult test to pass for SQL Server and my solution was the result_type check: result_type == TDS_DONE_RESULT before returning SUCCESS.  If you can send me a log file from Sybase I'll parse the hex and compare it with SQL server... possibly the fix will need to be something in tds_process_tokens.

freddy77:
I think I got the problem. Was more dbnextrow implementation. I merged 2 changeset and now your initial patch to fix the problem don't break rpc test.
Could you post a new pull request for the fix with a better commit message? The intial message at pull request #84 was pretty good.
Still running all tests but AppVeyour is passing and all 3 tests that had problems are passing with all servers so I don't expect surprises.

moogah:
Fantastic!  Do you want me to include the fix in the new PR or just the unit tests?

freddy77:
Just the fix. I already merged the unit tests (that's why AppVeyor master is failing). Did some minor changes. I think I can close this PR.

freddy77:
Closing, waiting for fix PR

PR #89, opened 2016-12-05
Batch statements without a select are not processed correctly.

moogah:

Signed-off-by: Jeff Farr jefarr@wayfair.com

### DIFF
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -1767,6 +1767,8 @@ _dbresults(DBPROCESS * dbproc)
 					dbproc->dbresults_state = _DB_RES_NEXT_RESULT;
 					if (done_flags & TDS_DONE_ERROR)
 						return FAIL;
+					if (result_type == TDS_DONE_RESULT)
+						return SUCCEED;
 					break;
 
 				case _DB_RES_RESULTSET_EMPTY:


freddy77:
Merged! I used this PR comment as the comment.

moogah:
In further testing, we've found a problem with this change.  SQL Server returns an empty rowset at the beginning and end of a BEGIN END block.  This is breaking backwards compatibility for TRY-CATCH, IF-THEN and possibly other statements. 

I've been parsing the hex more closely, looking for an identifier that separates these extra rowsets from ones that contain useful data but don't have a solution prepared yet.

moogah:
After more review, I have to recommend that this be taken out of the mainline until a solution can be found.  I don't yet see a clear identifier in the hex that would distinguish these additional rowsets and allow them to be filtered at the driver level.

freddy77:
However the tests you sent seems to respect proprietary behaviour.
I think we should sort out this regression as an improvement instead of just rolling back.
Can you open an issue?

moogah:
Sure!  I agree that it is better to have the more expected behavior and look for how to exclude the extra rowsets.  I'm not sure how much sprint work will be prioritized for me to work on this, but I can put together a test that demonstrates the issue.

PR #93, opened 2016-12-27
ODBC utf-16 when v 5.0 dts detected for sybase connections

mckaygerhard:
Sybase connections need utf-16 for correctly support of native unicode characters, this patch set by default this in odbc connections if are not specified and forceit if protocol v5.0 are choosend, due sybase seems only supports 5.0 assumes need of utf-16, on sybae AnyWhere unicode charsets are default


freddy77:
This patch just break some tests and don't fix nothing.
The client charset has nothing to do with server encoding.

Issue #105, opened 2017-03-09
Should `dbsettime` override the per-connection timeout settings?


attilahorvath:
I am working on adding per-connection timeout settings to TinyTDS, a Ruby gem which uses this library (https://github.com/rails-sqlserver/tiny_tds/pull/348).

I came across this issue that if we first assign a per-connection timeout value with `dbsetopt(connection, DBSETTIME, timeout, 0)` then later on call `dbsettime(timeout)`, this new timeout setting will override all the previously set timeouts for existing connections. It's pretty clearly happening in the implementation of `dbsettime` here: https://github.com/FreeTDS/freetds/blob/fc820490336c50d5c175d2a15327383256add4c9/src/dblib/dblib.c#L4105-L4123.

I was just wondering if this is the intended behaviour. I assumed that the global timeout setting should only be used for connections that don't have a unique timeout set for them.

If this is indeed a bug, I'd be happy to submit a patch to fix this.

freddy77:
Hi, was trying to check Microsoft documentation but apparently they removed everything about dblib.
I would say this is undefined behaviour at the moment.
Who win between global and local is not defined.
dtsettime (function) has to change all connection for compatibility with Sybase (which does not support dbsetopt(DBSETTIME)).
I think however that not changing the overridden ones make sense, Sybase application won't use dbsetopt(DBSETTIME) so the compatibility is retained for such apps.
Patch welcome.

attilahorvath:
I've submitted a PR for this: https://github.com/FreeTDS/freetds/pull/107. The Travis build is failing for some reason though (worked for me just fine locally).

freddy77:
Merged PR.

PR #107, opened 2017-03-10
Do not override per-connection timeout settings in dbsettime

attilahorvath:
As per the discussion in https://github.com/FreeTDS/freetds/issues/105, calling `dbsettime(timeout)` should not override per-connection timeout settings previously set by `dbsetopt(connection, DBSETTIME, timeout, 0)`.

### DIFF
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -4107,6 +4107,7 @@ dbsettime(int seconds)
 {
 	TDSSOCKET **tds;
 	int i;
+	DBPROCESS *dbproc;
 	tdsdump_log(TDS_DBG_FUNC, "dbsettime(%d)\n", seconds);
 
 	tds_mutex_lock(&dblib_mutex);
@@ -4114,8 +4115,11 @@ dbsettime(int seconds)
 	
 	tds = g_dblib_ctx.connection_list;
 	for (i = 0; i <  TDS_MAX_CONN; i++) {
-		if (tds[i])
-			tds[i]->query_timeout = seconds;
+		if (tds[i]) {
+			dbproc = (DBPROCESS *) tds_get_parent(tds[i]);
+			if (!dbisopt(dbproc, DBSETTIME, 0))
+				tds[i]->query_timeout = seconds;
+		}
 	}
 	
 	tds_mutex_unlock(&dblib_mutex);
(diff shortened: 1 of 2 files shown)


attilahorvath:
Not sure why the Travis build is failing. From the log it looks like it can't connect to the database for some reason.

src/dblib/unittests/timeout.c
```diff
@@ -187,6 +179,9 @@ test(int per_process)
 			fprintf(stderr, "Failed: dbsetopt(..., DBSETTIME, \"%d\")\n", timeout_seconds);
 			exit(1);
 		}
+
+		/* Verify setting the global timeout won't override the per-process timeout value */
+		dbsettime(120);
```
freddy77: Is this fails people will wait for quite long, I would use a 35 (query take 30 seconds).

src/dblib/unittests/timeout.c
```diff
@@ -162,14 +162,6 @@ test(int per_process)
...
 	}
 
-	dbsettime(60);
-
-	/* Verify setting a global timeout won't affect custom per-process timeouts */
-	if (dbproc->tcp_socket->query_timeout != timeout_seconds) {
-		printf("dbsettime() should not change per-process timeouts\n");
-		exit(-1);
-	}
-
```
freddy77: This hunk is removing an hunk added in the first patch. Can you remove from both?

src/dblib/dblib.c
```diff
@@ -4107,15 +4107,19 @@ dbsettime(int seconds)
...
 
 	tds_mutex_lock(&dblib_mutex);
 	g_dblib_ctx.query_timeout = seconds;
 	
 	tds = g_dblib_ctx.connection_list;
 	for (i = 0; i <  TDS_MAX_CONN; i++) {
-		if (tds[i])
-			tds[i]->query_timeout = seconds;
+		if (tds[i]) {
+			dbproc = (DBPROCESS *) tds[i]->parent;
```
freddy77: I would use tds_get_parent here

src/dblib/unittests/timeout.c
attilahorvath: Fair enough, changed it to 35.

src/dblib/unittests/timeout.c
attilahorvath: I've rebased the commit, so it should be gone.

src/dblib/dblib.c
attilahorvath: Haven't encountered that function, changed it now.

freddy77:
Merged

Issue #110, opened 2017-04-21
encryption = off treated like encryption = request in the freetds.conf file in login.c


gjcarrette:
There is a bug in login.c that causes encryption = off to be treated as encryption = request
whenever HAVE_OPENSSL is defined.  This is easily seen by using the tsql command
with TDSDUMPCONFIG and TDSDUMP defined and a Microsoft SQL Server that is TLS capable.
When you see tls.c mentioned in the TDSDUMP file you know it was not turned off.

Here is my test, where I edit the config file to have the encryption directive either missing (default)
or set to one of the 3 legal values, off, request, require. The strace helps validate what
config files were really used in the test.

```
  rm -f /tmp/freetds.check.*; \
  env FREETDSCONF=$HOME/freetds.conf \
      TDSDUMPCONFIG=/tmp/freetds.check.config.txt TDSDUMP=/tmp/freetds.check.log.txt \
      strace -f -s 128 -o /tmp/freetds.check.strace.txt /usr/local/bin/tsql -S JDBC -U guest -P sybase < /dev/null
```

freddy77:
Patch sound good. However I think the default should then be changed to request to not break lot of mssql configurations. Not sure what will happen to Sybase.

freddy77:
Merged, thanks

PR #113, opened 2017-04-28
Changes to the TDS_ENCRYPTION_LEVEL enum

hankinsoft:
Changes to the TDS_ENCRYPTION_LEVEL enum to match the values specified in https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-TDS/%5BMS-TDS%5D-160714.pdf. Precursor checkin with plans to add dblib option for specifying crypt level.

### DIFF
--- a/src/tds/login.c
+++ b/src/tds/login.c
@@ -1128,9 +1128,9 @@ tds71_do_login(TDSSOCKET * tds, TDSLOGIN* login)
 	/* encryption */
 #if !defined(HAVE_GNUTLS) && !defined(HAVE_OPENSSL)
 	/* not supported */
-	tds_put_byte(tds, 2);
+	tds_put_byte(tds, TDS_ENCRYPTION_NOT_SUPPORTED);
 #else
-	tds_put_byte(tds, login->encryption_level >= TDS_ENCRYPTION_REQUIRE ? 1 : 0);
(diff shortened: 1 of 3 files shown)


hankinsoft:
Note that I kept backwards compatibility. Previously TDS_ENCRYPTION_REQUIRE mapped to TDS_ENCRYPTION_NOT_SUPPORTED (0x02) and calling tds_config_encryption with TDS_STR_ENCRYPTION_REQUIRE actually sets the not supported flag. I think relates to issue #110.

hankinsoft:
Closing until #110/#114 are committed. Then I will resubmit.

src/tds/login.c
```diff
@@ -1128,9 +1128,9 @@ tds71_do_login(TDSSOCKET * tds, TDSLOGIN* login)
 	/* encryption */
 #if !defined(HAVE_GNUTLS) && !defined(HAVE_OPENSSL)
 	/* not supported */
-	tds_put_byte(tds, 2);
+	tds_put_byte(tds, TDS_ENCRYPTION_NOT_SUPPORTED);
```
freddy77: This patch confuses the configuration option with the protocol constants.

freddy77:
The two set of constant have not much to do.

PR #114, opened 2017-04-28
Fix the bug where encryption = off was treated like encryption = request

gjcarrette:
Also change the default to encryption = request so that existing users of encryption
will not see encryption turned off.

For https://github.com/FreeTDS/freetds/issues/110

This was the simple way to fix this problem, just be inspecting the obvious intention of the code
and without reference to outside reference documents. 

Let me explain the source to the bug and why I fixed it the way I did. A customer upgraded
to the new freetds package, but built it from source under new conditions where HAVE_OPENSSL
was defined. But the previous freetds package was configured without openssl devel libraries present.

Then the behavior of his system suddenly changed, due to the extra cost of negotiating
the TSL handshake on login.  To fix this I read the documentation and edited /etc/freetds.conf
to turn encryption = off; But that did not do anything.

The fix in the code for login.c was obvious. Treat encryption = off the same way as it would be treated if the software had not been built with HAVE_OPENSSL defined.

Now my pull requests appears to conflict with another pull requests for the same issue which makes
reference to documentation I have not read in detail. But I think if he changes his request
to also respond the comment about avoiding changes to default behavior, and
also updating the documentation, then the other request might address the issue fully.

### DIFF
--- a/src/tds/config.c
+++ b/src/tds/config.c
@@ -466,10 +466,10 @@ tds_config_boolean(const char *option, const char *value, TDSLOGIN *login)
 static void
 tds_config_encryption(const char * value, TDSLOGIN * login)
 {
-	TDS_ENCRYPTION_LEVEL lvl = TDS_ENCRYPTION_REQUEST;
+	TDS_ENCRYPTION_LEVEL lvl = TDS_ENCRYPTION_OFF;
 
 	if (!strcasecmp(value, TDS_STR_ENCRYPTION_OFF))
-		lvl = TDS_ENCRYPTION_OFF;
+		;
 	else if (!strcasecmp(value, TDS_STR_ENCRYPTION_REQUEST))
 		lvl = TDS_ENCRYPTION_REQUEST;
 	else if (!strcasecmp(value, TDS_STR_ENCRYPTION_REQUIRE))
--- a/src/tds/config.c
+++ b/src/tds/config.c
@@ -466,10 +466,10 @@ tds_config_boolean(const char *option, const char *value, TDSLOGIN *login)
 static void
 tds_config_encryption(const char * value, TDSLOGIN * login)
 {
-	TDS_ENCRYPTION_LEVEL lvl = TDS_ENCRYPTION_OFF;
+	TDS_ENCRYPTION_LEVEL lvl = TDS_ENCRYPTION_REQUEST;
 
 	if (!strcasecmp(value, TDS_STR_ENCRYPTION_OFF))
-		;
+		lvl = TDS_ENCRYPTION_OFF;
 	else if (!strcasecmp(value, TDS_STR_ENCRYPTION_REQUEST))
 		lvl = TDS_ENCRYPTION_REQUEST;
 	else if (!strcasecmp(value, TDS_STR_ENCRYPTION_REQUIRE))
--- a/doc/freetds.conf.5.in
+++ b/doc/freetds.conf.5.in
@@ -102,9 +102,9 @@ no
 .It encryption
 .Bl -tag -compact
 .It Em off
-disables encryption (default)
+disables encryption
 .It Em request
-use if available
+use if available (default)
 .It Em required
 allow encrypted connections only
 .El
--- a/doc/userguide.sgml
+++ b/doc/userguide.sgml
@@ -1154,8 +1154,8 @@ This is the name of the database container in the server you are connecting to.<
 						<row>
 							<entry><literal>encryption</></entry>
 							<entry>off/request/require</entry>
-							<entry>off</entry>
-							<entry>Specify if encryption is desired. Supported for Microsoft servers. <symbol>off</> disables encryption (only if needed); <symbol>request</> means use if available; <symbol>require</> means create and allow encrypted connections only.</entry>
+							<entry>request</entry>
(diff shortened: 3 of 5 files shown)


freddy77:
Make sense. However I would like to see what happens on Sybase servers setting request. Are they set by default to handle encryption?
Not sure also if the override code in config.c can deal with the new default.

src/tds/config.c
```diff
@@ -466,10 +466,10 @@ tds_config_boolean(const char *option, const char *value, TDSLOGIN *login)
 static void
 tds_config_encryption(const char * value, TDSLOGIN * login)
 {
-	TDS_ENCRYPTION_LEVEL lvl = TDS_ENCRYPTION_OFF;
+	TDS_ENCRYPTION_LEVEL lvl = TDS_ENCRYPTION_REQUEST;
 
 	if (!strcasecmp(value, TDS_STR_ENCRYPTION_OFF))
-		;
+		lvl = TDS_ENCRYPTION_OFF;
```
freddy77: Did you test this? This is not changing the default at all.
The default should be set in tds_alloc_login.

freddy77:
I wrote a patch to change the default to request for MSSQL (which currently is the same for MSSQL) while for compatibility Sybase uses off (Sybase 15 installation would fail if changed).
Can you update your patch? Should require just login.c changes (I changed also some constant using some mnemonic).

src/tds/config.c
gjcarrette: I recall single-stepping through this function. But I think perhaps my changes here are needed, plus I need to change tds_alloc_login also, when there is no encryption configuration.

It is line 469 where I change the default` lvl to TDS_ENCRYPTION_REQUEST `as you suggested for backward compatibility. Then if the user specified `encryption = off `in the freetds.conf file this line of code 472 is needed, whereas in the past it would not be needed, but would certainly have been OK to have written.  With the change in line 469 one could, if one was obsessive about optimizations, remove line of code 474, since this would be a NOOP. However, I choose not to optimize code to the extent that it is confusing so leave this line of code in place: `lvl = TDS_ENCRYPTION_REQUEST`.

freddy77: https://github.com/FreeTDS/freetds, master branch. The specific patch is https://github.com/FreeTDS/freetds/commit/f0400f56a34ce13fd3ece82cdcc21177a0715a8d. I had to change default to off for Sybase and request for MSSQL.
I think the ";" statement was there to avoid some warnings from compilers that don't understand lvl is always set when accessed at the end. The initial lvl initialization is not the default. Note that lvl is set in all paths except where the string value is OFF (in this case lvl is set to OFF).

gjcarrette: The thing that fooled me was that git grep encryption_level did not find tds_alloc_login because the value was being serendipitously set to 0 by the TEST_MALLOC macro using calloc.
I updated my branch, also updating the documentation, and ran this test to make sure I knew exactly what config files were being used by each test:

```
      rm -f /tmp/freetds.check.*
      env FREETDSCONF=$HOME/freetds.conf \
          TDSDUMPCONFIG=/tmp/freetds.check.config.txt TDSDUMP=/tmp/freetds.check.log.txt \
          strace -f -s 128 -o /tmp/freetds.check.strace.txt make check
```

I did not include the patch of @freddy77 yet however. Need to understand how gitlab works better to find it. Stay tuned...

gjcarrette:
I am *still* working on this. For one thing I discovered that the comment "The default should be set in tds_alloc_login." is not true. Since login.c already had  this code everywhere it was needed:
```
       if (login->encryption_level == TDS_ENCRYPTION_DEFAULT)
                login->encryption_level = TDS_ENCRYPTION_REQUEST;
```

So the documentation was wrong about encryption default being off. 
Since TDS_ENCRYPTION_DEFAULT = 0 the calloc in mem.c already had login encryption_level = 0 = default.

Still doing more testing. My previous change to config.c was a case of a change that was a NOOP, but did no harm, and where I did actually do testing, and found default behavior was correct. So many if defined and if and even "?" version of if involved here.

gjcarrette:
What do you think about this code in login.c function `tds_connect` around line 521. The `else if` is useless because it duplicates the previous `if` does it not? I found this during testing and single stepping with gdb...

```
        if (IS_TDS71_PLUS(tds->conn)) {
                erc = tds71_do_login(tds, login);
                db_selected = 1;
        } else if (IS_TDS7_PLUS(tds->conn)) {
                erc = tds7_send_login(tds, login);
                db_selected = 1;
        } else {
                tds->out_flag = TDS_LOGIN;
                erc = tds_send_login(tds, login);
        }
```

freddy77:
The
```
       if (login->encryption_level == TDS_ENCRYPTION_DEFAULT)
                login->encryption_level = TDS_ENCRYPTION_REQUEST;
```
was added after the comment "The default should be set in tds_alloc_login." to handle Sybase. Yes, documentation should be changed too, but was conflicting with your patch. Now the default is OFF for Sybase and REQUEST for MSSQL. Now change in tds_alloc_login is not needed.

gjcarrette:
This got so messy with some many intermediate changes and comments that I have created a
fresh branch and am going to submit that as a merge request. Functionally it goes back to the
origin change to login.c but with a more detailed comment based on current understanding
and with changes to the documentation to reflect current reality.

PR #116, opened 2017-05-02
Fix the bug where encryption = off was treated like encryption = request

gjcarrette:
in the freetds.conf Also updated the documentation to describe
how the encryption default depends on the tds version. Existing users
of encryption should be ok, since TDS_ENCRYPTION_DEFAULT always becomes
either TDS_ENCRYPTION_OFF or TDS_ENCRYPTION_REQUEST.

For https://github.com/FreeTDS/freetds/issues/110


freddy77:
Merged, thanks

PR #123, opened 2017-06-02
In `ct_options`, set the tds_argsize to 1 for CS_OPT_ARITHIGNORE.

meissnem:
I am encountering errors when calling `ct_options` with `CS_OPT_ARITHIGNORE`.  Sybase server 15.7 is returning:

    Sybase Message: Severity 18, Error 3805, State 1

    The token datastream length was not correct. This is an internal
    protocol error.

It seems that `tds_argsize` is set to 4 when my server is expecting it to be 1. This patch changes that.

### DIFF
--- a/src/ctlib/ct.c
+++ b/src/ctlib/ct.c
@@ -3539,8 +3539,8 @@ ct_options(CS_CONNECTION * con, CS_INT action, CS_INT option, CS_VOID * param, C
 			if (action == CS_SET)
 				return CS_FAIL;
 		}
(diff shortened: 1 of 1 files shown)


freddy77:
Did some tests. There's only a byte however the encoding is a bit different. Simply the boolean is passed to the server (using one byte) using the same TDS_OPT_ARITHIGNOREON option.
I think is possible to just add CS_OPT_ARITHIGNORE to tds_bool_option_map (not tested).

freddy77:
Yes, using just the boolean handling code fix the issue.
Thanks.

meissnem:
Thanks, @freddy77 -- I appreciate you taking care of this!

PR #126, opened 2017-06-17
tds: Escape left brace in regex in encodings.pl

RavuAlHemio:
Since Perl v5.22.0, unescaped left braces ('{') in regexes are deprecated; since Perl v5.26.0, they are forbidden in most cases. Escape such a case in encodings.pl using a backslash.

(I am currently hunting down a case of memory corruption in `fisql`, but found this change necessary for FreeTDS to build.)


freddy77:
Merged, thanks

PR #127, opened 2017-06-17
dblib buffer length management fixes

RavuAlHemio:
This pair of commits fixes a few bugs with buffer length management in `dbsprhead`.


freddy77:
Merged the first patch.
About the second looking at the log it does not make much sense to print a buffer that can be uninitialized as used for the output. Maybe changing %s to %p wound be a better solution.

RavuAlHemio:
Yup, that makes sense. I've updated the PR accordingly.

freddy77:
Merged, thanks

PR #132, opened 2017-07-05
Fix out-of-tree build on MINGW.

larskanis:
[MSYS2](http://msys2.org) builds packages out-of-tree. This currently fails because `odbc_w.def` can not be found in the build directory.

Therefore I added this patch in the build script for a FreeTDS build script here: https://github.com/Alexpux/MINGW-packages/pull/2671


freddy77:
Merged, thanks

PR #141, opened 2017-08-11
configure.ac: support "without" versions of three ./configure flags.

orlitzky:
The `--with-iodbc`, `--with-unixodbc`, and `--with-odbc-nodm` flags all use
the standard AC_ARG_WITH macro. The resulting `./configure` script
accepts e.g. `--without-iodbc` to indicate that the user does not want
to enable iODBC, and in place of a path the word "no" is placed into
the `$with_iodbc` variable. The current `configure.ac` doesn't handle that
and instead treats "no" as a path. This commit updates `configure.ac` to
ignore "no" as the value of those three flags.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -546,7 +546,7 @@ esac
 
 AC_ARG_WITH(iodbc,
 AS_HELP_STRING([--with-iodbc=DIR], [build odbc driver against iODBC in DIR]))
-if test "$with_iodbc"; then
+if test "x$with_iodbc" != "x" -a "x$with_iodbc" != "xno"; then
 	if echo "$with_iodbc" | grep -v '^/'; then
 		with_iodbc="$PWD/$with_iodbc"
 	fi
@@ -563,7 +563,7 @@ fi
 
 AC_ARG_WITH(unixodbc,
 AS_HELP_STRING([--with-unixodbc=DIR], [build odbc driver against unixODBC in DIR]))
-if test "$with_unixodbc"; then
+if test "x$with_unixodbc" != "x" -a "x$with_unixodbc" != "xno"; then
 	if echo "$with_unixodbc" | grep -v '^/'; then
 		with_unixodbc="$PWD/$with_unixodbc"
 	fi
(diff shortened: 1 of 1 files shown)


orlitzky:
To test: before this commit, `--without-iodbc` would crash the configure script because it couldn't find the iODBC headers. Afterwards, the configure script shouldn't even check (it does the same thing as if the `$with_iodbc` was undefined).

configure.ac
```diff
@@ -546,7 +546,7 @@ esac
 
 AC_ARG_WITH(iodbc,
 AS_HELP_STRING([--with-iodbc=DIR], [build odbc driver against iODBC in DIR]))
-if test "$with_iodbc"; then
+if test "x$with_iodbc" != "x" -a "x$with_iodbc" != "xno"; then
```
freddy77: Not a big fun of the x prefix. I would use just
`if test "$with_iodbc" != "" -a "$with_iodbc" != "no"; then`
same below

configure.ac
```diff
@@ -588,7 +588,7 @@ fi
 
 AC_ARG_WITH(odbc_nodm,
 AS_HELP_STRING([--with-odbc-nodm=DIR], [build odbc using headers in DIR/include]))
-if test "x$with_odbc_nodb" != "x" -a "x$with_odbc_nodm" != "xno"; then
+if test "x$with_odbc_nodm" != "x" -a "x$with_odbc_nodm" != "xno"; then
```
freddy77: Can you squash this fix into the first commit?

orlitzky:
Sure, I squashed them. I'm no expert on this, but if I recall correctly, the silly "x" thing is for compatibility with some old shells where "test" was not a builtin function, but instead a binary located in e.g. `/bin`. Attempting to use that binary for an empty-string comparison would cause something weird to happen.

I'll see if I can find a reference, just for my own curiosity.

freddy77:
One reason for the "x" could be values starting with "-".

By the way. Merged.

Thanks.

orlitzky:
Yeah, the best I could find was this Stack Overflow response:

https://stackoverflow.com/questions/6852612/bash-test-for-empty-string-with-x

It mentions the leading hyphen, too, but I don't know if any modern shells or `/bin/test` programs have that problem. I'm happy to drop the "x" if you want.

PR #145, opened 2017-08-13
configure.ac: autodetect iODBC when --with-iodbc=yes.

orlitzky:
This is the simplest thing that works and doesn't interfere with the rest of the build system... just looking for some initial feeback.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -577,6 +577,9 @@ fi
 AC_ARG_WITH(unixodbc,
 AS_HELP_STRING([--with-unixodbc=DIR], [build odbc driver against unixODBC in DIR]))
 if test "x$with_unixodbc" != "x" -a "x$with_unixodbc" != "xno"; then
+	if test "x$with_iodbc" != "x" -a "x$with_iodbc" != "xno"; then
+		AC_ERROR([choose at most one of --with-iodbc or --with-unixodbc])
+	fi
 	if echo "$with_unixodbc" | grep -v '^/'; then
 		with_unixodbc="$PWD/$with_unixodbc"
 	fi
--- a/doc/userguide.sgml
+++ b/doc/userguide.sgml
@@ -395,13 +395,26 @@ Cf. <link linkend="withOdbcNoDM"><option>--with-odbc-nodm</></link>.</para>
 				
 				<variablelist id="tab.ODBC.Driver.Managers"><title>ODBC Driver Managers</title>
 					<varlistentry>
-						<term><Option>--with-iodbc=<replaceable>DIR</>
+					        <term><Option>--with-iodbc<replaceable>
+							</Option></term>
+					        <term><Option>--with-iodbc=<replaceable>DIR</>
 							</Option></term>
 						<term><Option>--with-unixodbc=<replaceable>DIR</>
 							</Option></term>
-						<listitem><para>Specify directory of &iODBC; or &unixODBC; support, and use it as the Driver Manager.
-								As of version 0.62, the ODBC Driver Manager is detected by <command>configure</>, so use this parameter only if yours is installed in a nonstandard path.
-								(Requires &iODBC; or &unixODBC; to have already been installed.)</para>
+                                                <listitem>
+						  <para>
+						    Specify a particular ODBC driver manager and the directory in which it is installed.
+						    The <Option>--with-iodbc</Option> form chooses &iODBC; as the driver manager, and <Option>--with-unixodbc</Option> specifies &unixODBC; as the driver manager.
+						    The directory argument is required for <Option>--with-unixodbc</Option>, but may be omitted for <Option>--with-iodbc</Option>; pkg-config will be used to find your &iODBC; installation if the directory is omitted.
+						    Typical directory arguments are <filename>/usr</filename> and <filename>/usr/local</filename>.
+						  </para>
+						  <para>
+						    So long as either &iODBC; or &unixODBC; are installed, the build system will detect your driver manager by default.
+						    As a result, these options are only needed if you wish to override the default behavior.
+						  </para>
+						  <para>
+						    It is an error to specify both <Option>--with-iodbc</Option> and <Option>--with-unixodbc</Option>.
+						  </para>
 							</listitem>
 						</varlistentry>
 						<varlistentry id="withOdbcNoDM">


configure.ac
```diff
@@ -546,7 +546,28 @@ esac
...
 AS_HELP_STRING([--with-iodbc=DIR], [build odbc driver against iODBC in DIR]))
-if test "x$with_iodbc" != "x" -a "x$with_iodbc" != "xno"; then
+
+if test "x$with_iodbc" = "xyes"; then
+	# The user asked for iodbc support, but didn't tell us
+	# where to look. So, we'll try to get the info from
+	# pkg-config. If we can't, we'll display an error.
+	PKG_CHECK_MODULES(IODBC, libiodbc)
+	CPPFLAGS="$CPPFLAGS -DIODBC $IODBC_CFLAGS"
+	LDFLAGS="$LDFLAGS $IODBC_LIBS"
```
freddy77: This should not be added to LDFLAGS so you don't link odbc library to dblib and ctlib.
You should add to ODBCLIB instead (and not use -liodbc in ODBCLIB).

configure.ac
```diff
@@ -546,7 +546,28 @@ esac
...
 AC_ARG_WITH(iodbc,
 AS_HELP_STRING([--with-iodbc=DIR], [build odbc driver against iODBC in DIR]))
-if test "x$with_iodbc" != "x" -a "x$with_iodbc" != "xno"; then
+
+if test "x$with_iodbc" = "xyes"; then
+	# The user asked for iodbc support, but didn't tell us
+	# where to look. So, we'll try to get the info from
+	# pkg-config. If we can't, we'll display an error.
+	PKG_CHECK_MODULES(IODBC, libiodbc)
+	CPPFLAGS="$CPPFLAGS -DIODBC $IODBC_CFLAGS"
```
freddy77: Not a problem adding to global but I would add IODBC_CFLAGS to ODBC_INC for compatibility.

configure.ac
```diff
@@ -546,7 +546,28 @@ esac
...
+	# The IODBC_LIBS variable can contain some big-L linker flags,
+	# and the ODBCLIB variable is substituted in a few places where
+	# those flags are inappropriate. So for now, we just mimic the
+	# case below and define ODBCLIB ourselves. A refactoring could
+	# use another variable, such as ODBC_LDFLAGS, instead.
+	#
+	# The ODBC_INC variable is only substituted into *_CPPFLAGS,
+	# which is there $ODBC_CFLAGS is supposed to go anyway.
+	ODBCLIB="-liodbc"
+	ODBC_INC="$ODBC_CFLAGS"
```
freddy77: I think there's a typo, IODBC_CFLAGS (same on comment)

configure.ac
```diff
@@ -546,7 +546,28 @@ esac
 
 AC_ARG_WITH(iodbc,
 AS_HELP_STRING([--with-iodbc=DIR], [build odbc driver against iODBC in DIR]))
-if test "x$with_iodbc" != "x" -a "x$with_iodbc" != "xno"; then
+
+if test "x$with_iodbc" = "xyes"; then
```
freddy77: Didn't check but maybe we would need to change documentation too to state that directory is optional.

configure.ac
orlitzky: It is, now that I've fixed the typo.

configure.ac
orlitzky: Fixed, thanks.

configure.ac
orlitzky: This is what the comment is talking about.. neither choice is really good. The `IODBC_LIBS` variable can contain both the `-lfoo` and `-L/usr/foo` type of linker flags.

By placing `IODBC_LIBS` in `LDFLAGS`, we pass some extra `-l` flags to every linker command.

On the other hand, the `ODBCLIB` variable appears in some places where `-L/usr/foo` flags would cause problems:

* `bsqlodbc_LDADD	= $(ODBCLIB) $(ODBCNODMLIB)` in src/apps/Makefile.am
* `LIBS		=	$(ODBCLIB)...` in src/odbc/unittests/Makefile.am

What we really want is a set of LDFLAGS that only applies to the ODBC stuff, so that's why I mentioned a potential refactoring to involve an `ODBC_LDFLAGS` variable.

configure.ac
orlitzky: Yep you're right, but first I have to figure out how to build the docs.

configure.ac
freddy77: Yes, you are right, was a bit of a mess.
I updated and now at least Makefiles uses just LDFLAGS options, removed LIBS/LDADD/LIBADD (at least for ODBC).
Renamed ODBCLIB to ODBC_LDFLAGS.

freddy77:
Added a commit on top of yours to update the code, see https://github.com/FreeTDS/freetds/tree/orlitzky-iodbc-pkgconfig. Feel free to ignore and sent your version. I think a single commit is enough.

orlitzky:
Fantastic, thanks -- I was going to attempt to do something similar but I wasn't looking forward to it. I added another commit that tries a similar trick for `--with-unixodbc=yes`. I don't personally have a use for that case, but consistency is good.

That just leaves the documentation update.

orlitzky:
After I walked away for a few minutes, I realized it was probably better to perform the same "did we find unixODBC?" test in both branches, so the new commit adds that. I can squash it if you prefer the new test.

configure.ac
```diff
@@ -562,6 +576,20 @@ fi
...
+
+if test "x$with_unixodbc" = "xyes"; then
+	# The user asked for unixodbc support, but didn't tell us
+	# where to look. We'll try to get the installation prefix from
+	# the odbc_config executable, and if we can't, we'll display
+	# an error. We require odbc_config to be in $PATH.
+	ODBC_INC=`odbc_config --include-prefix`
+	test -r "$ODBC_INC/sql.h" || AC_ERROR([sql.h not found])
+	test -r "$ODBC_INC/sqlext.h" || AC_ERROR([sqlext.h not found])
+	with_unixodbc=`odbc_config --prefix`
```
freddy77: This looks a bit wrong. with_unixodbc should contain the option passed to configure, why you are changing it?

configure.ac
```diff
@@ -562,6 +576,20 @@ fi
 
 AC_ARG_WITH(unixodbc,
 AS_HELP_STRING([--with-unixodbc=DIR], [build odbc driver against unixODBC in DIR]))
+
+if test "x$with_unixodbc" = "xyes"; then
+	# The user asked for unixodbc support, but didn't tell us
+	# where to look. We'll try to get the installation prefix from
+	# the odbc_config executable, and if we can't, we'll display
+	# an error. We require odbc_config to be in $PATH.
+	ODBC_INC=`odbc_config --include-prefix`
```
freddy77: this is just the directory, should also add the -I option.
I honestly won't change unixODBC.
I hope in a future they'll add support for pkg-config too.

freddy77:
Merged iODBC patch.

configure.ac
orlitzky: That code is only reached with `--with-unixodbc=yes`. The idea is that, if I pass `--with-unixodbc=yes`, then it will try to use `odbc_config` to find `sql.h` and `sqlext.h`. If it succeeds, then it will change `$with_odbc` from `yes` to `/usr/`, and then continue as if I had specified `--with-unixodbc=/usr` in the first place.

Anyway I don't need this patch so I'm fine if you omit it, thanks for all your help.

I'll look into the documentation updates over the weekend.

freddy77:
What you want to change is probably the source for http://www.freetds.org/userguide/config.htm, specifically the part with --with-iodbc=DIR.
You can simply open the giant doc/userguide.sgml file with a text editor, look for "--with-iodbc" and add some text to that part, kind of "If DIR is not specified ..."
Once you sent the patch I can compile and if there's problem or small changes to make I can propose an update.
The documentation is in SGML, the parent of HTML and XML so for instance "&" needs to be changed in "&amp;" and space rules are similar to HTML (multiple spaces are handled as a single one except some particular cases).

freddy77:
Oh... I did some minor changes just now to the file so update your git before editing.

orlitzky:
I just force-pushed this branch, so all that should remain is the update to the documentation. It wasn't hard to get it to build, I just needed to install the docbook stuff and run:

```
$ export DOCBOOK_DSL="/usr/share/sgml/docbook/dsssl-stylesheets-1.79/html/docbook.dsl"
$ make
```

doc/userguide.sgml
```diff
@@ -395,13 +395,19 @@ Cf. <link linkend="withOdbcNoDM"><option>--with-odbc-nodm</></link>.</para>
...
-								As of version 0.62, the ODBC Driver Manager is detected by <command>configure</>, so use this parameter only if yours is installed in a nonstandard path.
-								(Requires &iODBC; or &unixODBC; to have already been installed.)</para>
+                                                <listitem>
+						  <para>
+						    Specify a particular ODBC driver manager and the directory in which it is installed. The <Option>--with-iodbc</Option> form chooses &iODBC; as the driver manager, and <Option>--with-unixodbc</Option> specifies &unixODBC; as the driver manager. The directory argument is required for <Option>--with-unixodbc</Option>, but may be omitted for <Option>--with-iodbc</Option>; pkg-config will be used to find your &iODBC; installation if the directory is omitted. Typical directory arguments are <filename>/usr</filename> and <filename>/usr/local</filename>.
+						  </para>
+						  <para>
+						    So long as either &iODBC; or &unixODBC; are installed, the build system will detect your driver manager by default. As a result, these options are only needed if you wish to override the default behavior.
```
freddy77: Same split comment

orlitzky: Sorry, what would you like me to change in those lines?

Unrelated: is it valid to specify both `--with-iodbc` and `--with-unixodbc`? If not, I could add that to the documentation as well (and make it an error in `configure.ac`).

freddy77: Was suggesting instead of

  <xxx>
    Sentence. Sentence. Sentence.

something like

  <xxx>
   Sentence.
   Sentence.
   Sentence.

this to make future diff smaller.

as --with-iodbc and --with-unixodbc should force one of the other user should not specify both. Currently not enforced.

orlitzky:
No problem, I split each sentence onto its own line, and also documented the fact that `--with-iodbc` and `--with-unixodbc` are mutually exclusive. In a new commit, I made the ./configure script bail out if the user passes both flags.

freddy77:
Merged, thanks. Rebase and removed a small error in the file.

PR #148, opened 2017-08-31
CTLIB: ct_options handle TEXTSIZE command

Kokan:
In the ct_options missed the CS_OPT_TEXTSIZE configuration option, the rest of the code had already done it. It would be nice to set this property via the API not just via configuration file.


freddy77:
Merged with some updates (spaces and "i" size).

Thanks

PR #158, opened 2017-09-30
Update CMakeLists.txt to define INSTALL files, targets... in Windows Systems

jalegido:
When someone is compiling freetds in windows and use CMake, he can't install the developer files in an install folder.
This pull request adds necesary install target to CMakeLists.txt to install *.h, *.lib, *.exe and *.dll in next directory structure:
/install_folder
/install_folder/include/*.h
/install_folder/include/freetds/version.h
/install_folder/lib/*.lib
/install_folder/bin/*.exe *.dll

This pull install all targets defined in CMakelists.txt, but I'm not sure if all *.lib must be exported.
Neither, I'm not sure if all headers files must be exported or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -474,15 +474,15 @@ set(FREETDS_PUBLIC_INCLUDE
         ${CMAKE_SOURCE_DIR}/include/sybdb.h
 		${CMAKE_SOURCE_DIR}/include/sybfront.h 
         ${CMAKE_SOURCE_DIR}/include/syberror.h
-		${CMAKE_SOURCE_DIR}/include/odbcss.h
-		${CMAKE_BINARY_DIR}/include/tds_sysdep_public.h)
+		${CMAKE_SOURCE_DIR}/include/odbcss.h)
 
 set(FREETDS_PUBLIC_INCLUDE_2 
 		${CMAKE_BINARY_DIR}/include/freetds/version.h)
 
 set(FREETDS_CONF_FILES 
 		${CMAKE_SOURCE_DIR}/freetds.conf
-		${CMAKE_SOURCE_DIR}/locales.conf)
+		${CMAKE_SOURCE_DIR}/locales.conf
+		${CMAKE_SOURCE_DIR}/src/pool/pool.conf)
 		
 install(FILES ${FREETDS_PUBLIC_INCLUDE} DESTINATION include)
 install(FILES ${FREETDS_PUBLIC_INCLUDE_2} DESTINATION include/freetds)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -461,3 +461,30 @@ configure_file(${CMAKE_SOURCE_DIR}/include/freetds/version.h.in ${CMAKE_BINARY_D
 if(NOT EXISTS "${CMAKE_SOURCE_DIR}/configure")
 	configure_file(${CMAKE_SOURCE_DIR}/src/odbc/version.rc.in ${CMAKE_SOURCE_DIR}/src/odbc/version.rc)
 endif(NOT EXISTS "${CMAKE_SOURCE_DIR}/configure")
+
+
+if(WIN32)
+set(FREETDS_PUBLIC_INCLUDE 
+		${CMAKE_SOURCE_DIR}/include/bkpublic.h
+        ${CMAKE_SOURCE_DIR}/include/cspublic.h
+        ${CMAKE_SOURCE_DIR}/include/cstypes.h
+        ${CMAKE_SOURCE_DIR}/include/ctpublic.h
+		${CMAKE_SOURCE_DIR}/include/sqldb.h
+		${CMAKE_SOURCE_DIR}/include/sqlfront.h
+        ${CMAKE_SOURCE_DIR}/include/sybdb.h
+		${CMAKE_SOURCE_DIR}/include/sybfront.h 
+        ${CMAKE_SOURCE_DIR}/include/syberror.h
+		${CMAKE_SOURCE_DIR}/include/odbcss.h
+		${CMAKE_BINARY_DIR}/include/tds_sysdep_public.h)
+
+set(FREETDS_PUBLIC_INCLUDE_2 
+		${CMAKE_BINARY_DIR}/include/freetds/version.h)
+
+set(FREETDS_CONF_FILES 
+		${CMAKE_SOURCE_DIR}/freetds.conf
+		${CMAKE_SOURCE_DIR}/locales.conf)
+		
+install(FILES ${FREETDS_PUBLIC_INCLUDE} DESTINATION include)
+install(FILES ${FREETDS_PUBLIC_INCLUDE_2} DESTINATION include/freetds)
+install(FILES ${FREETDS_CONF_FILES} DESTINATION etc)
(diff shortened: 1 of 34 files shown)


CMakeLists.txt
```diff
@@ -461,3 +461,30 @@ configure_file(${CMAKE_SOURCE_DIR}/include/freetds/version.h.in ${CMAKE_BINARY_D
...
+		${CMAKE_SOURCE_DIR}/include/sqldb.h
+		${CMAKE_SOURCE_DIR}/include/sqlfront.h
+        ${CMAKE_SOURCE_DIR}/include/sybdb.h
+		${CMAKE_SOURCE_DIR}/include/sybfront.h 
+        ${CMAKE_SOURCE_DIR}/include/syberror.h
+		${CMAKE_SOURCE_DIR}/include/odbcss.h
+		${CMAKE_BINARY_DIR}/include/tds_sysdep_public.h)
+
+set(FREETDS_PUBLIC_INCLUDE_2 
+		${CMAKE_BINARY_DIR}/include/freetds/version.h)
```
freddy77: why you want to install that file?

CMakeLists.txt
```diff
@@ -461,3 +461,30 @@ configure_file(${CMAKE_SOURCE_DIR}/include/freetds/version.h.in ${CMAKE_BINARY_D
...
+
+set(FREETDS_CONF_FILES 
+		${CMAKE_SOURCE_DIR}/freetds.conf
+		${CMAKE_SOURCE_DIR}/locales.conf)
+		
+install(FILES ${FREETDS_PUBLIC_INCLUDE} DESTINATION include)
+install(FILES ${FREETDS_PUBLIC_INCLUDE_2} DESTINATION include/freetds)
+install(FILES ${FREETDS_CONF_FILES} DESTINATION etc)
```
freddy77: not sure what should be the destination of these. I think the code looks at c:\

src/apps/CMakeLists.txt
```diff
@@ -15,3 +15,12 @@ endif(WIN32)
 
 add_executable(bsqlodbc bsqlodbc.c)
 target_link_libraries(bsqlodbc tdsodbc replacements tdsutils ${libs})
+
+if(WIN32)
+INSTALL(TARGETS tsql bsqlodbc
```
freddy77: quite lot of repetitions. Maybe defining a install_win32_targets would help

CMakeLists.txt
jalegido: I would like verify if freetds version is greater than a specific value.
I can use TDS_VERSION_MAJOR, TDS_VERSION_MINOR,... to assert the version when code is compiled.

Regarding tds_sysdep_public.h I saw it in Makefile.am and I added it. But as I said in previous comment, I only use ctlib so I don't know if all headers included and libraries are necesary.

CMakeLists.txt
jalegido: I'm using this commands to build it in Windows.

set OPENSSL_ROOT_DIR=C:\develop\openssl-1.1.0e\x64
set OUT_RELEASE=C:\develop\tds\x64

cmake .. -G "Visual Studio 14 2015 Win64" -DWITH_OPENSSL=ON -DENABLE_KRB5=ON -DOPENSSL_ROOT_DIR=%OPENSSL_ROOT_DIR% -DCMAKE_INSTALL_PREFIX:PATH=%OUT_RELEASE%
cmake --build . --config Release --target install

With this command, I'm defining that all results must be in c:\develop\tds\x64. With the directories hierarchy defined in each CMakeLists.txt

${FREETDS_PUBLIC_INCLUDE},... is pointing to developer files (${CMAKE_SOURCE_DIR}/include...) and with cmake install clause it should be copied to ${CMAKE_INSTALL_PREFIX} set in the user command

src/apps/CMakeLists.txt
jalegido: There are a lot of repetitions. I would like to do only one install clause, but (At least as far as I know) cmake doesn't allow to install targets outside each CMakelists.txt.

freddy77: was thinking about defining a "install_targets_win32" macro in the main CMakeLists.txt to avoid all directory repetition so you could use something like install_targets_win32(tdsl bsqlodbc)

freddy77: why only installing on win32 ?

jalegido: For me it's ok defining "install_targets_xxxx" and not write repetitions in each CMakeLists.txt
I just added win32, because it is the platform where I'm testing. 
But the install process can be used in all platforms.

CMakeLists.txt
```diff
@@ -461,3 +461,30 @@ configure_file(${CMAKE_SOURCE_DIR}/include/freetds/version.h.in ${CMAKE_BINARY_D
...
+        ${CMAKE_SOURCE_DIR}/include/syberror.h
+		${CMAKE_SOURCE_DIR}/include/odbcss.h
+		${CMAKE_BINARY_DIR}/include/tds_sysdep_public.h)
+
+set(FREETDS_PUBLIC_INCLUDE_2 
+		${CMAKE_BINARY_DIR}/include/freetds/version.h)
+
+set(FREETDS_CONF_FILES 
+		${CMAKE_SOURCE_DIR}/freetds.conf
+		${CMAKE_SOURCE_DIR}/locales.conf)
```
freddy77: standard installed file include pool.conf too

src/odbc/CMakeLists.txt
```diff
@@ -43,3 +43,12 @@ add_library(tdsodbc_static STATIC
 if (NOT WIN32)
 	set_target_properties(tdsodbc_static PROPERTIES COMPILE_FLAGS -fPIC)
 endif()
+
+if(WIN32)
+INSTALL(TARGETS tdsodbc tdsodbc_static
```
freddy77: the static library should not be installed, not intended to be reused

src/tds/CMakeLists.txt
```diff
@@ -47,3 +47,12 @@ add_library(tds STATIC
 if (NOT WIN32)
 	set_target_properties(tds PROPERTIES COMPILE_FLAGS -fPIC)
 endif()
+
+if(WIN32)
+INSTALL(TARGETS tds
```
freddy77: not intended to be installed and reused, not sure if you need to compile with static dblib/ctlib

src/utils/CMakeLists.txt
```diff
@@ -19,3 +19,12 @@ add_library(tdsutils STATIC
 if (NOT WIN32)
 	set_target_properties(tdsutils PROPERTIES COMPILE_FLAGS -fPIC)
 endif()
+
+if(WIN32)
+INSTALL(TARGETS tdsutils
```
freddy77: about this and replacement is not clear if we want to be used

jalegido:
I've removed all internal libraries and tds_sysdep_public.h from the installed files; pool.conf is added to installed files too.

Regarding install_win32_targets, I don't know how to add subdirectories targets.

If we update install information to linux system, it's necesary follow the directory rules for this system. As bin, lib, etc, include... Maybe these paths are different from Windows system.

freddy77:
Marged. Some minimal space changes.

PR #168, opened 2017-10-28
updated to use markdown

EvanCarroll:
I upgraded all the text docs internally to use markdown and render properly on github.


EvanCarroll:
Just doc patches, but I do have a question on COPYING vs COPYING_LIB. It's confusing to distribute both GPL and LGPL. AFAIK, LGPL is a superset: that is to say the license is simply more liberal. If that's the case, my suggestion is just to remove the GPLv2 and distribute the LGPLv2. Actually, I would prefer to license under the copy-left, but being that it's not the case if this is an LGPLv2 library then there is no reason to distribute a more restrictive license with less binding.

freddy77:
Sorry for late reply.
Many files are referenced in different part (like spec file) which would need updates. I prefer project to build correctly than having a nice view on github.
There's no rationale on the commit.
The COPYING files I think should stay that format for legal issues.
About LGPL and GPL we have both, the rule of thumb is that all libraries are LGPL while programs are GPL.

freddy77:
ping?

freddy77:
No reply in months

EvanCarroll:
I'm not sure how to update the build system, sorry no experience with automake. Also GNU hosts the Markdown versions of the license too, and I'm not sure how it could be argued that the rendering makes a difference.

I think if anything was precarious for the project it would be this,

> About LGPL and GPL we have both, the rule of thumb is that all libraries are LGPL while programs are GPL.

I don't think that a rule of thumb would be enforceable in court. If the project is listed as LGPL (and it is), I'm not sure how you could bind someone to GPL for just the executable. Sounds murky.

freddy77:
Yes, maybe you can convert also licensing but I don't see the point of changing to a more editable format something that needs to be kept verbatim.

There's no legal issue, every file has his own copyright specified clearly, the rule of thumb is just to avoid a long answer.

About the build basically different files refers to the file you changed so needs to be updated so the original file should be generate again or the name changed. Usually the names are referred in Makefile.am, some are implicit in automake, I don't know how to change but I can find out.

About the rationale you should state the reason of this change in your commit. Is just style? Easier to change (from web) ? Formats can be changed easily (like producing html files) ?

freddy77:
Sorry took so long, finally merged

EvanCarroll:
@freddy77 shows unmerged to me.

freddy77:
@EvanCarroll look at the code, it was merged

PR #170, opened 2017-11-17
dblib: fix rpc blob data output

fandrieu:
The dblib API to get data from an RPC output parameter, `dbretdata` doesn't handle BLOB types, while the same works for result sets with `dbdata`.
As both functions seem to handle the same kind of data (from `tds_put_data`), here a proposed fix that brings `dbretdata` up to date.
There's 3 commits in this branch: the first one is a minimal patch, then a cleaner one copies all logic and finally a common internal function is added to avoid duplication.

This works in my limited testing and it seems unlikely the current code (a pointer to column_data) is relied upon, thanks for looking into this !

### DIFF
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -4735,7 +4735,8 @@ dbretdata(DBPROCESS * dbproc, int retnum)
 		return NULL;
 
 	column = param_info->columns[retnum - 1];
-	/* FIXME blob are stored is different way */
+	if (is_blob_col(column))
+		return (BYTE *) ((TDSBLOB *) column->column_data)->textvalue;
 	return (BYTE *) column->column_data;
 }
 
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -61,6 +61,7 @@
 #include <dblib.h>
 
 static RETCODE _dbresults(DBPROCESS * dbproc);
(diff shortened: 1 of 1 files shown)


freddy77:
Looks good, do you have an example query?

fandrieu:
I tested this using the the [cli tool](https://github.com/fandrieu/freetds/commits/freerpc) I wrote that runs rpc queries :)
Using a TDS 7.2 connection, this should output "test":
`freerpc sp_executesql -p "set @a='test'" -p "@a text out" -t char -op a -S mssql`
But this 4 random chars:
`freerpc sp_executesql -p "set @a='test'" -p "@a text out" -t text -op a -S mssql`

A query with an output parameter of type text/ntext/image set in dbrpcparam should reproduce this, the output length is correct but the data is not.

freddy77:
Merged! Nice idea using sp_executesql. Maybe less nice to call as public interface, better to execute directly code.

fandrieu:
Thanks ! Got it, better come up with a unit test.
[Here's one](https://github.com/fandrieu/freetds/commit/3c053087fbe8ee577ad0e25eaf59ced02c1b6510) that uses a varchar(max) custom procedure.

freddy77:
Can't you extend the old one? I already added a test in rpc.c (see master). Could you open another PR?

fandrieu:
I wasn't sure how to use `sql_cmd`, will do another PR.

PR #171, opened 2017-11-17
Fix blob(4) size for converted text

fandrieu:
In my tests, data sent as NTEXT using TDS 7.0 (varint 4) appears to be truncated in half, due to an incorrect size set in tds_generic_put.
This patch ensures the actual column size it set to the converted data length.

### DIFF
--- a/src/dblib/unittests/.gitignore
+++ b/src/dblib/unittests/.gitignore
@@ -28,6 +28,7 @@ t0016.err
 t0017.err
 t0017.out
 rpc
+rpc70
 dbmorecmds
 bcp
 thread
--- a/src/dblib/unittests/CMakeLists.txt
+++ b/src/dblib/unittests/CMakeLists.txt
@@ -2,7 +2,7 @@ add_library(d_common STATIC common.c common.h)
 
 foreach(target t0001 t0002 t0003 t0004 t0005 t0006 t0007 t0008 t0009
 	t0011 t0012 t0013 t0014 t0015 t0016 t0017 t0018 t0019 t0020
-	t0021 t0022 t0023 rpc dbmorecmds bcp thread text_buffer
+	t0021 t0022 t0023 rpc rpc70 dbmorecmds bcp thread text_buffer
 	done_handling timeout hang null null2 setnull numeric pending
 	cancel spid canquery batch_stmt_ins_sel batch_stmt_ins_upd bcp_getl)
 	add_executable(d_${target} EXCLUDE_FROM_ALL ${target}.c)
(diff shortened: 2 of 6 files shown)


freddy77:
Can you provide a test? Which API are you using?

fandrieu:
It's with the dblib API, using the same tool as the other PR, this sould print 4:
`freerpc sp_executesql -p "select len(@a)" -p "@a char(10)" -t char -p TEST -V 7.0 -S mssql`
`freerpc sp_executesql -p "select len(@a)" -p "@a char(10)" -t text -p TEST -V 7.0 -S mssql`

But this 2:
`freerpc sp_executesql -p "select len(@a)" -p "@a char(10)" -t ntext -p TEST -V 7.0 -S mssql`

(it all works with tds >  7.2 and varint_size 8)

fandrieu:
I added a unit test, without the fix it should fail with errno 8016 "Data type 0x63 has an invalid data length"

freddy77:
I'll have a look. Wondering why to truncate BLOBs sending. There are some limit but there's no BLOB(xx), BLOBs do not support PRECISION. But on network level the type info contains a length specification.

I also tried and with TDS 7.1 on output it either give an error stating that type is wrong (from mssql 2005) or the output specification is ignored (before mssql 2005). That is BLOB output is not supported before 7.2 (note, this is not related to this PR, this is about input, not output).

fandrieu:
Sorry the blob(4) notation I came up with wasn't clear. It was meant to refer to the "real" SYBNTEXT as opposed to XSYBNVARCHAR of column size 8.

In my limited understanding the former is sql "NTEXT" (usable with TDS 7.0, but not for output parameters :)) and the later is sql "NVARCHAR(MAX)" (usable with TDS 7.2, including for output parameters). As the varint_size 8 types are automatically used by FreeTDS if available I wanted to be more precise...

But the important thing is you understood me :)  Thanks for looking into this !

freddy77:
I'm considering this patch instead https://github.com/FreeTDS/freetds/commit/e5330ac8820ae5eaaa1ce063ce1f341d832b9208. Basically tell all blobs can be maximum length that would result not truncating the conversion (or better truncate at the physical limit).
I run all tests in the environment I use and the patch is working, all servers do not complains.
Can you try with your tests and see if works too?

fandrieu:
Yes this works with my tests.
I actually started with a fix like this but wasn't sure of the consequences so I went with a less intrusive approach...
Anyway, that's way nicer, thanks !

freddy77:
Good. I merged that patch (with an updated comment).

PR #173, opened 2017-11-19
dblib rpc test: blob output procedure

fandrieu:
As a follow up to #170, here's a test for rpc blob output that uses a custom procedure.


freddy77:
Not clear why this test is better than the other, could you not just change the "@a text out" to "@a varchar(max) out" ?

freddy77:
Sorry, maybe you changed the test after I said in a previous comment sp_executesql should not be used. I was referring to daily and user level usage more than test usage (where hackish code is acceptable as long as maintainable). This test started to fails as (as you wrote in a comment) is dependent on "text size" option.

Why didn't you just use a "set @txt=replicate(cast('pad-len-10' as varchar(max)),6000)+'test123'" in the store procedure?

fandrieu:
Yes exactly, I hadn't seen the RPC unittest and liked the "temp procedure" approach, I though it could be nice to follow the same path and avoid introducing `sp_executesql` into the mix.
So there's no real improvement in this new test, except maybe that it actually fetches a large string.

Oh, I didn't use `replicate` because I didn't thought of that...maybe writing C made me forget about the nice things in  higher level languages :) Anyway nice idea, a lot cleaner !

fandrieu:
I updated the branch with a call to `set textsize` to ensure the test works and the use of `replicate` in the procedure. Don't hesitate to close this PR if you think this change is not worth it.

freddy77:
Yes, I think better to close this ticket, the test is already doing its job with sp_executesql and by the way varchar(max) requires already TDS 7.2 which means actually quite old server.

PR #174, opened 2017-11-24
Support for nossl builds

jpgimenez:
Based on @ramiro's work I added support for nossl builds for windows, this is needed for packages like pymssql or ctds that want to be distributed statically linked against FreeTDS but did the same with openssl could be an issue.
Right now the libs with and without ssl support are all bundled on the same zip file, but could be distributed as separated files if that is preferred.


freddy77:
Not a big fun of having TLS disabled nowadays but I understand there are cases you want to do it.
The series is really in a bad shape. Lot of fix/revert stuff, I think should just be a single commit.
You are doing everything in "after_test" but I think would be better to do the build in "build_script" instead.
I would also follow the same directory schema for ssl/nossl.

jpgimenez:
Thanks @freddy77 for review it, yes I know that it looks bad, I can re-do it in a single commit and move it to build_script, but before doing it I have one question... what do you mean with "same directory schema"?

lib/ssl/dynamic
lib/ssl/static
lib/nossl/dynamic
lib/nossl/static

or what?
again, I can generate two separate zip files if preferred.

freddy77:
Yes, was referring at that directory schema you reported. I just changed using "make install" way, the schema is now changed but there's no more manual mkdir/copy to do.

jpgimenez:
I'll re-open this PR after implement a better install and some clean-up.

PR #178, opened 2017-11-29
Support TLS versions greater than TLSv1 when using OpenSSL

tschafer:
Currently, FreeTDS supports TLSv1.1 and TLSv1.2 only when used with GnuTLS. This adds support for TLSv1.1 and TLSv1.2 with OpenSSL.

### DIFF
--- a/src/tds/tls.c
+++ b/src/tds/tls.c
@@ -94,7 +94,7 @@ BIO_get_data(const BIO *b)
 {
 	return b->ptr;
 }
-#define TLS_client_method TLSv1_client_method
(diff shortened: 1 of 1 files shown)


freddy77:
Patch looks good, I would just use parenthesis around the define, like

`#define DEFAULT_OPENSSL_CTX_OPTIONS (SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3)`

tschafer:
Good point. Here you go...

freddy77:
Merged, tried even with an old server and it works.

PR #181, opened 2017-12-13
add nossl build on appveyor

jpgimenez:
generate a separate package without ssl support, to allow pymssql windows packages to be build statically linked without ssl, its in a separate package to simplify the installation and to keep packages small


freddy77:
Merged, thanks.

PR #184, opened 2018-01-04
dblib: fix TDS_DONE_RESULT empty rowsets:

### LINKED ISSUE #175: Empty recordsets with num_cols > 0
This is a follow up to #156 

After executing SET-like statements, SQL Server returns a "DONE" token.
This produces an empty recordset in FreeTDS, but it can be identified because it doesn't have any column.

But this doesn't work after a regular recordset has been received: the number of columns reported stays the same, it's not reset to 0.

So, with the following query:
```sql
set arithabort on
select 1
set arithabort on
select 2
```
I would expect 4 results, with 0, 1, 0, 1 columns
But the actual number of columns reported is 0, 1, 1, 1.

To be clear, I [wrote a test](https://github.com/FreeTDS/freetds/commit/32f307d3f0c0d9358e93eca6caadd14a358e59bb) showing this using dblib.

fandrieu:
Fix #175 

Change introduced by 34e1f69 can produce an incorrect resultset.
This patch resets the internal result structure in this case.

### DIFF
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -1760,8 +1760,10 @@ _dbresults(DBPROCESS * dbproc)
 					dbproc->dbresults_state = _DB_RES_NEXT_RESULT;
 					if (done_flags & TDS_DONE_ERROR)
 						return FAIL;
-					if (result_type == TDS_DONE_RESULT)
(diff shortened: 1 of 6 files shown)


freddy77:
Marged, thanks.

PR #197, opened 2018-03-09
Memory leak in tls.c

alexloney:
OpenSSL documentation for[ SSL_get_peer_certificate](https://www.openssl.org/docs/manmaster/man3/SSL_get_peer_certificate.html) states:

> The reference count of the X509 object is incremented by one, so that it will not be destroyed when the session containing the peer certificate is freed.

Meaning that, when the certificate hostname is checked, OpenSSL keeps a reference to that certificate each time, so each connection that's made will leak more and more memory.

Since the cert is not used after the hostname check, I've added a X509_free() call that will free the certificate.


freddy77:
Merged, thanks.

PR #198, opened 2018-03-14
Using system root CAs when establishing a connection.

alexloney:
The issue with the method that existed already (the call to [SSL_CTX_set_default_verify_paths](https://www.openssl.org/docs/man1.1.0/ssl/SSL_CTX_set_default_verify_paths.html)) is that it will:

> SSL_CTX_set_default_verify_paths() specifies that the default locations from which CA certificates are loaded should be used. There is one default directory and one default file. The default CA certificates directory is called "certs" in the default OpenSSL directory. Alternatively the SSL_CERT_DIR environment variable can be defined to override this location. The default CA certificates file is called "cert.pem" in the default OpenSSL directory. Alternatively the SSL_CERT_FILE environment variable can be defined to override this location.

So, from my understanding, it sets it to a default directory and a default file of certs.pem. Which would work fine if /etc/ssl/certs (default directory on my installation) contained a single consolidated certs.pem file, but it does not. It instead contains many .pem files, one for each root CA.

Furthermore, when using FreeTDS on Windows, there is no directory of root certificates, instead all of the certificates are managed via the Microsoft Management Console through the Certificates store. As far as I'm aware, OpenSSL does not know of this system wide certificate store (it's the same store that Chrome and other browsers use to get their trusted root certificates).

Finally, you should be able to see this lack of knowledge by FreeTDS/OpenSSL on either Linux or Windows by setting `encryption=require` and `ca file=system` in the freetds config for the server. This will force OpenSSL to establish a secure connection and attempt to use the system CA files. However, the connection should fail (at least it has failed in my build) because it does not successfully load the root CAs, and thus does not successfully verify the connection.

The change that I've proposed adds an additional function of `tds_update_cert_store_with_root_cas` to tls.c that will do two different things when called:

1. On Windows, it will make a call to open the [CertOpenSystemStore](https://msdn.microsoft.com/es-es/library/windows/desktop/aa376560(v=vs.85).aspx), when open, it will loop through all root certificates in the store and add them to the OpenSSL context's root store, providing a chain of trust for OpenSSL to verify with.
2. On Linux, it will get the default directory for OpenSSL to read certificates from via a call to X509_get_default_cert_dir_env or X509_get_default_cert_dir, it will then loop through every .pem file in that directory, read them in as X509 certs, and add them to the OpenSSL context.

The result is that, on both Windows and Linux, when using OpenSSL, the certificate chain will be able to be verified when `encryption=require` and `ca file=system` are both set.

### DIFF
--- a/src/tds/tls.c
+++ b/src/tds/tls.c
@@ -1035,6 +1035,7 @@ tds_ssl_init(TDSSOCKET *tds)
 		tls_msg = "checking hostname";
 		if (!cert || !check_hostname(cert, tds_dstr_cstr(&tds->login->server_host_name)))
 			goto cleanup;
+		X509_free(cert);
 	}
 
 	tdsdump_log(TDS_DBG_INFO1, "handshake succeeded!!\n");
--- a/src/tds/tls.c
+++ b/src/tds/tls.c
@@ -584,12 +584,12 @@ int tds_update_cert_store_with_root_cas(SSL_CTX *ctx)
 	X509 *x509 = NULL;
 #ifdef WIN32
 	HCERTSTORE hStore = NULL;
-    PCCERT_CONTEXT pContext = NULL;
+	PCCERT_CONTEXT pContext = NULL;
 #else
-    FILE *f = NULL;
-    const char *path = NULL;
-	char *fullPath = NULL;
-    DIR *dirp = NULL;
+	FILE *f = NULL;
+	const char *path = NULL;
+		char *fullPath = NULL;
+	DIR *dirp = NULL;
 	struct dirent entry;
 	struct dirent *endp;
 	struct stat st;
@@ -599,8 +599,8 @@ int tds_update_cert_store_with_root_cas(SSL_CTX *ctx)
 
 #ifdef WIN32
 	hStore = CertOpenSystemStore(NULL, "ROOT");
-    if (!hStore)
-        goto err;
+	if (!hStore)
+		goto err;
 
 	while (pContext = CertEnumCertificatesInStore(hStore, pContext))
 	{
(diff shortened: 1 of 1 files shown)


freddy77:
Can you update the change on last master? Looks like the indentation of the code is a bit wrong.
Beside that looks good.

alexloney:
Due to my editor displaying tabs as 4 spaces, I didn't notice that tabs and spaces had gotten mixed in the source file. I've committed an update that replaced the spaces with tabs to keep the tab based formatting.

freddy77:
How did you test these patches? On Windows they do not even compile.

alexloney:
What issue are you seeing when you attempt to compile on Windows? I have successfully compiled with Microsoft Visual Studio 2012.

freddy77:
See https://ci.appveyor.com/project/FreeTDS/freetds/build/1033, basically crypt32 is not linked in.
Beside that, how did you test it?

alexloney:
Ah yes, CertOpenSystemStore function on Windows is in the Crypt32.lib file. I have that linked into my tds project, but it probably needs to be added to the CMakeLists.txt somewhere to automatically get added when building on Windows?

To test the changes, this is what I did:

I first noticed that, when a SQL Server does not have an SSL certificate assigned to it, it will generate a self-signed certificate on-the-fly to encrypt traffic, however because this is a self-signed certificate, it doesn't prevent a MITM attack from happening, as the MITM could generate their own self-signed certificate and sit between the client and server.

So I then had an SSL certificate generated from a trusted root CA.

On SQL Server, I went to the computer running SQL Server and loaded the new certificate into Microsoft Management Console, followed by going to Server Configuration Manager and selecting that as the certificate for the SQL Server itself.

After researching a bit inside of FreeTDS, I found out that the only way to have it actually verify the chain of trust is to have the configuration set a TDS version >= 7.1, an encryption of "required" and a CA file of "system". So after doing that, it successfully connects to the SQL Server.

Now, that proves a positive case, but what I'm really concerned about is the negative case, proving that an invalid cert (e.g. a MITM attack with a self-signed cert) will fail verification in OpenSSL. So I went to the client computer and deleted the root CA file that the servers SSL certificate chains up to, then attempted a connection as before. Without that root CA file in the system, OpenSSL failed to connect to the server (as expected). I then restored the root CA file.

To test on Linux, it was a similar process, only I'm not root on the server that it's testing on, so I compiled a version of FreeTDS that would read from the wrong directory and verify that the connection attempt failed (because it doesn't have the root CA file loaded), restored the correct file path, and verified that it successfully connected to the server.

PR #200, opened 2018-03-22
optimize the padding blanks, fixed the hotspot

hex2tan:
[In a case following:
The destlen is too big such like 10Mb, but the srclen/len is just a few bytes, will loop almost 10 * 1024 * 1024 times to pad the blank. And then this will be a man-made hotspot.

My story...
In my rencently project, our team use freetds as a client to connect SQL Server on linux platform. We have a VARCHAR(MAX) type column in database, this column was used to store string, the length is range from 0 bytes to 10MB, mostly data were a few bytes.The function dbcollen() return 2GB when the type is VARCHAR(MAX), we just roughly truncated to 10Mb. When we got a short length data which was VARCHAR(MAX) and call the dbbind with a specified bind size(10MB), desttype(STRINGBIND).This will loop too much.
The following is Intel VTune Amplifier result.
freetds_host_var0

freetds_host_var1

freetds_host_var2

### DIFF
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -7500,7 +7500,7 @@ copy_data_to_host_var(DBPROCESS * dbproc, TDS_SERVER_TYPE srctype, const BYTE * 
 					if (srclen < destlen) {
 						memset(dest + srclen, ' ', destlen - srclen - 1);
 					}
-					dest[i] = '\0';
+					dest[destlen - 1] = '\0';
 					break;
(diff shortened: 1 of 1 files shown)


freddy77:
I would merge the 2 commits and remote the "fixed the hotspot" from the title, hard to understand if you don't use VTune daily.
Anyway, tests are failing.

freddy77:
Fixed and merged

PR #204, opened 2018-04-01
Fix build failure with LibreSSL 2.7

Sp1l:
LibreSSL 2.7 added OpenSSL 1.1 API leading to build failures

See also: https://bugs.freebsd.org/226911
Signed-off-by: Bernard Spil <brnrd@FreeBSD.org>

### DIFF
--- a/src/tds/sec_negotiate_openssl.h
+++ b/src/tds/sec_negotiate_openssl.h
@@ -37,6 +37,11 @@
 #error HAVE_OPENSSL not defines, this file should not be included
 #endif
 
+#if defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2070000fL
+#undef OPENSSL_VERSION_NUMBER
+#define OPENSSL_VERSION_NUMBER 0x1000107fL
+#endif
+
 static inline const BIGNUM*
 rsa_get_n(const RSA *rsa)
 {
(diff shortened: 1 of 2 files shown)


src/tds/tls.c
```diff
@@ -75,6 +75,15 @@
 #define SSL_PTR ptr
 #else
 
+#ifdef LIBRESSL_VERSION_NUMBER
+#if LIBRESSL_VERSION_NUMBER < 0x2070000FL
+static pthread_mutex_t *openssllocks;
```
freddy77: where is this variable used?

src/tds/sec_negotiate_openssl.h
```diff
@@ -37,6 +37,11 @@
 #error HAVE_OPENSSL not defines, this file should not be included
 #endif
 
+#if defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2070000fL
+#undef OPENSSL_VERSION_NUMBER
+#define OPENSSL_VERSION_NUMBER 0x1000107fL
+#endif
+
```
freddy77: looks like these lines are reused in the below file, should not they be in an header, like freetds/tls.h?

freddy77:
Did you see my comments?

src/tds/sec_negotiate_openssl.h
Sp1l: That sounds correct... Let me fix the patch

PR #252, opened 2019-01-26
docker install doc

belegnar:


INSTALL
```diff
@@ -26,3 +26,15 @@ Brief Instructions
...
 
    For any other information please see the Users Guide.
+
+Docker installation (for Debian-based images)
+
+    RUN apt install -y git automake autoconf libtool make gcc perl gettext unixodbc-dev
+    RUN git clone https://github.com/FreeTDS/freetds
+    RUN cd freetds
+    WORKDIR freetds
+    RUN git checkout Branch-1_00
```
freddy77: Now should probably be Branch-1_1

INSTALL
```diff
@@ -26,3 +26,15 @@ Brief Instructions
    with unixODBC 2.2.11.
 
    For any other information please see the Users Guide.
+
+Docker installation (for Debian-based images)
+
+    RUN apt install -y git automake autoconf libtool make gcc perl gettext unixodbc-dev
```
freddy77: What is `RUN` ? Means run a command or is a Docker prompt? Usually I prefer the classic `$` for normal user and `#` for root

INSTALL
```diff
@@ -26,3 +26,15 @@ Brief Instructions
...
+    RUN git clone https://github.com/FreeTDS/freetds
+    RUN cd freetds
+    WORKDIR freetds
+    RUN git checkout Branch-1_00
+    RUN autoreconf -vfi
+    RUN ./autogen.sh --enable-libiconv --with-unixodbc=/usr --with-tdsver=7.1
```
freddy77: Please remove `--with-tdsver=7.1`, new version moved to autodetect which should be left as default

INSTALL
```diff
@@ -26,3 +26,15 @@ Brief Instructions
...
+    RUN make
+    RUN make install
```
freddy77: This make me think you are running all as root user inside the Docker container

INSTALL
freddy77: Not sure that you need `gettext`, can you check?
Would not be better to download a snapshot and avoid `libtool`, `automake`, `autoconf` and `perl`?

freddy77:
Closing, no replies in months

PR #258, opened 2019-02-13
CTLIB: ct_con_props() set/get client_charset

alltilla:
I wanted to set the client side character encoding from libdbi, but I could not find a proper setter/getter API for it. I saw, that similar options were set/got in ct_con_props() so I decided to add it there.


include/cspublic.h
```diff
@@ -248,6 +248,8 @@ enum
 #define CS_SEC_APPDEFINED CS_SEC_APPDEFINED
 	CS_STICKY_BINDS = 9151,
 #define CS_STICKY_BINDS CS_STICKY_BINDS
+	CS_CLIENTCHARSET = 9152,
```
freddy77: I think would be better as 9300, Sybase is using 9152 for CS_HAVE_CMD

freddy77:
Looks clean enough to be merged in next release.
Only comment is the number, can you update and move the constant after the 9300 one?

include/cspublic.h
alltilla: Sure thing, thanks for the quick reply :)

freddy77:
Merged, did couple of fixes

PR #259, opened 2019-02-14
make: look for userguide_desc.sgml in build dir

alltilla:
Building out-of-tree did not work because of missing the userguide_desc.sgml file in the source directory.

userguide_desc.sgml is generated to the build dir.
Previously make looked for userguide_desc.sgml in the source/doc directory, which was not a problem, when we are building in tree.
However, if we build out of tree it will not find the file.

Reproduction steps:
 1. cd to source root dir
 2. `env NOCONFIGURE=1 ./autogen.sh ; mkdir build ; cd build ; ../configure && make && echo Success`


freddy77:
Can you try just removing the first line:
`test -r $(abs_srcdir)/userguide_desc.sgml`
I'm getting other issues with your patch.

alltilla:
You are right, removing that one line is enough. Thanks!

freddy77:
Merged

PR #261, opened 2019-02-15
make: Add subdirs of src to DIST_SUBDIRS

alltilla:
Subdirectories of `src/` should get into the tarball when we `make dist`, no matter the configuration before.

I ran into this problem, when I configured with `--disable-pool --disable-server` options, and created a `make dist` tarball. When I untar'd the tarball, I tried to configure and build with the same flags, but it threw an error during configure:
`config.status: error: cannot find input file: 'src/server/Makefile.in'`

Reproduction command:
`env NOCONFIGURE=1 ./autogen.sh ; ./configure --disable-pool --disable-server && make dist ; tar xzvf freetds*.tar.gz && cd freetds* ; ./configure --disable-pool --disable-server && make && echo Success`

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -9,12 +9,17 @@ endif
 
 if INCPOOL
 SUBDIRS += server pool
-else !INCPOOL
+else
+DIST_SUBDIRS += server pool
 if INCSERVER
 SUBDIRS += server
+else
+DIST_SUBDIRS += server
 endif
 endif
 
 if INCAPPS
 SUBDIRS += apps
+else
+DIST_SUBDIRS += apps
 endif


src/Makefile.am
```diff
@@ -9,12 +9,17 @@ endif
 
 if INCPOOL
 SUBDIRS += server pool
-else !INCPOOL
```
freddy77: I think it would be easier to just initialize `DIST_SUBDIRS` as
`DIST_SUBDIRS = utils replacements tds ctlib dblib odbc server pool apps`

freddy77: It's working! Can I change the patch accordingly?

alltilla: Sure! Thank you for you kind help :)

freddy77:
Merge the fix, thanks

PR #262, opened 2019-02-19
tds.h: include socket.h

alltilla:
On AIX the `sys/socket.h` file is needed directly for values like `AF_INET` or `AF_INET6`.
There were 2 files which used these values, but did not include `socket.h`, and both of them had `tds.h` included, so I decided to put the include there.

The files are:
src/replacements/getaddrinfo.c
src/replacements/socketpair.c


freddy77:
Merged, thanks

alltilla:
Thanks!

PR #266, opened 2019-03-04
Fix bulk insert multibyte string support in column/table name

r313pp:
When column name (or table name in blob type case) has non ASCII characters, length of such string was calculated incorrectly.
In such cases bulk insert fails with:
```
Msg 4804, Level 17, State 1
        While reading current row from host, a premature end-of-message was encountered--an incoming data stream was interrupted when the server expected to see more data. The host program may have terminated. Ensure that you are using a supported client application programming interface (API).
```

To reproduce:
1. Use `client charset = UTF8` (with other charsets you might encounter other error (buffer overflow while converting column name - this error is not addressed hire)).
2. Create table with column that contains at least 1 character which is > 1 byte in UTF-8.
For example:
    * `¯\_(ツ)_/¯`. Current len 13, correct 9.
    * `ж`. Single Cyrillic letter. Current len 2, correct 1.
    * `Yдд`. One ASCII + 2 Cyrillic letters. Current len 5, correct 3.
3. `BULK INSERT` at least 1 row of any correct value in that column => ERROR.

MSSQL expects `UTF-16 length / 2` (actual length in letters), current implementation gives `client charset length`.

### DIFF
--- a/src/tds/bulk.c
+++ b/src/tds/bulk.c
@@ -736,7 +736,10 @@ tds7_bcp_send_colmetadata(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
 		TDSSTATICINSTREAM r;
 		TDSSTATICOUTSTREAM w;
 		/* table and column names are sysname(nvarchar(128)) which is max 128*2 bytes */
-		char sysname_buf[128*2];
+		/* table name can be in form of database_name.schema_name.table_name */
+		/* also database_name, schema_name, table_name can contain [] or "" as quotes and escaping of ]] or "" */
+		/* so we get (128 * 2 * 2 + 2) * 3 + 2 maximum */
+		char sysname_buf[(128 * 2 * 2 + 2) * 3 + 2];
 		int res;
 
 		bcpcol = bcpinfo->bindinfo->columns[i];
--- a/src/tds/bulk.c
+++ b/src/tds/bulk.c
@@ -732,15 +732,8 @@ tds7_bcp_send_colmetadata(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
 	tds_put_smallint(tds, num_cols);
 
 	for (i = 0; i < bcpinfo->bindinfo->num_cols; i++) {
-		size_t len, len_ucs2;
-		TDSSTATICINSTREAM r;
-		TDSSTATICOUTSTREAM w;
-		/* table and column names are sysname(nvarchar(128)) which is max 128*2 bytes */
-		/* table name can be in form of database_name.schema_name.table_name */
-		/* also database_name, schema_name, table_name can contain [] or "" as quotes and escaping of ]] or "" */
-		/* so we get (128 * 2 * 2 + 2) * 3 + 2 maximum */
-		char sysname_buf[(128 * 2 * 2 + 2) * 3 + 2];
-		int res;
+		size_t converted_len;
+		const char *converted_name;
 
 		bcpcol = bcpinfo->bindinfo->columns[i];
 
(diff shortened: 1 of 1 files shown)


src/tds/bulk.c
```diff
@@ -760,15 +768,26 @@ tds7_bcp_send_colmetadata(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
...
 		 */
 		if (is_blob_type(bcpcol->on_server.column_type)) {
-			/* FIXME support multibyte string */
 			len = tds_dstr_len(&bcpinfo->tablename);
-			TDS_PUT_SMALLINT(tds, len);
-			tds_put_string(tds, tds_dstr_cstr(&bcpinfo->tablename), len);
+			tds_staticin_stream_init(&r, tds_dstr_cstr(&bcpinfo->tablename), len);
+			tds_staticout_stream_init(&w, &sysname_buf, sizeof(sysname_buf));
+			res = tds_convert_stream(tds, tds->conn->char_convs[client2ucs2], to_server, &r.stream, &w.stream);
+			len_ucs2 = w.stream.buffer - sysname_buf;
```
freddy77: Maybe would be easier to use tds_convert_string/tds_convert_string_free ?

r313pp: I thought that something like `tds_convert_string` must exists, but not managed to find it.

r313pp: Rewritten using `tds_convert_string `.

src/tds/bulk.c
```diff
@@ -760,16 +761,28 @@ tds7_bcp_send_colmetadata(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
 		 * different from BCP format
 		 */
 		if (is_blob_type(bcpcol->on_server.column_type)) {
-			/* FIXME support multibyte string */
-			len = tds_dstr_len(&bcpinfo->tablename);
-			TDS_PUT_SMALLINT(tds, len);
-			tds_put_string(tds, tds_dstr_cstr(&bcpinfo->tablename), len);
+			converted_name = tds_convert_string(tds, tds->conn->char_convs[client2ucs2], tds_dstr_cstr(&bcpinfo->tablename), (int)tds_dstr_len(&bcpinfo->tablename), &converted_len);
+			if (!converted_name) {
+				return TDS_FAIL;
```
r313pp: Should I `tds_set_state` hire prior to returning?

src/tds/bulk.c
```diff
@@ -760,16 +761,28 @@ tds7_bcp_send_colmetadata(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
...
+			}
+
+			/* UTF-16 length is always size / 2 even for 4 byte letters (yes, 1 letter of length 2) */
+			TDS_PUT_SMALLINT(tds, converted_len / 2);
```
r313pp: Hire `TDS_PUT_SMALLINT` is used, but for column name length `tds_put_byte`.
Is this intentional?

src/tds/bulk.c
r313pp:
```
TDSRET
tds_bcp_start(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
{
	...

	if (IS_TDS7_PLUS(tds->conn))
		tds7_bcp_send_colmetadata(tds, bcpinfo);
	
	return TDS_SUCCESS;
}
```
`tds_bcp_start` when calling `tds7_bcp_send_colmetadata` does not check for errors.
Should I fix this?

freddy77: Good idea, not a regression here. Even better to write a better test for BCP and no-ascii encoding. But still not a regression here.

freddy77: As you probably will loose sync with the server is a good idea to disconnect.
I won't expect to happen, mostly would fail for memory errors.

freddy77:
merged

PR #276, opened 2019-03-29
Escape column names

Svelix:
We have a table in our project with a column named 'key'.
Since key is a reserved keyword this leads to errors when the column name is used without escaping it. 
The SQL that is created by using the defncopy tool did not escape column names so it lead to errors.

This pull-request changes defncopy so column names are escaped using square brackets. My C-skills are a little rusty though, so please check if I did everything right.


freddy77:
Patch make sense, some consideration:
- would be good to also support possible ']' in names;
- what would happen if a column name contain a comma ?
- not all servers support [] for quoting, old ASE ones only support "";
- (OT?) maybe for these utility some more relax "abort on memory error" should be considered.

Michoels:
What needs to be done before this can be merged?

freddy77:
Superceded

PR #277, opened 2019-04-03
Fix for accessing "SQL Database managed instance" on Azure

ThorstenEngel:
Dear Freddy,

recently I started evaluating an "Azure managed SQL Server (most recent: 12.0.2000.8)" which did not work from python (pymssql). I quickly learned, I'll have to manualy compile FreeTDS with OpenSSL. This worked, and I was able to connect with TSQL, but not via pymssql. I started debugging (TDSDUMPCONFIG) and found the only difference:

- tsql: server_name=host
- pymssql: server_name=host:1433

Googling around brought up "https://github.com/pymssql/pymssql/issues/330" to workarund the "Cannot open server "1433D"" problem. So it is suggested to use "**user@host**" instead of only "user". This workaround is even mentioned on "http://pymssql.org/en/stable/azure.html" and in Azure Docs as well.

**But this workaround did not work.** 

So I started analysing the sourcecode of pymssql, which looked okay to me (it passes "host:port") to FreeTDS. I analysed then FreeTDS and found "_config.c_" which does a lot with the server_name (e.g. stripping away the port and setting the port in the connection struct - _parse_server_name_for_port_). Further down (where my change is) it undoes all these changes and resets server_name to the data of the login. "1 || " Looks as well as someone put this in for debugging reasions...

I had to add the check for "TDS_DEF_SERVER" to not break the behaviour from TSQL. An alternative approach would be a strstr for ':'.

I hope that my fix is extremely helpful for all people that use pymssql and maybe even PHP with Azure SQL. 

With this fix applied:
- TSQL works (tested succesfully: onpremise MSSQL, Azure SQL managed)
- pymssql works (tested succesfully: onpremise MSSQL, Azure SQL managed)
- the username will be applied **without** @host (simply "user")

Warm regards from Munich

Thorsten

### DIFF
--- a/src/tds/config.c
+++ b/src/tds/config.c
@@ -701,7 +701,8 @@ tds_config_login(TDSLOGIN * connection, TDSLOGIN * login)
 	DSTR *res = &login->server_name;
 
 	if (!tds_dstr_isempty(&login->server_name)) {
-		if (1 || tds_dstr_isempty(&connection->server_name)) 
+        // overwrite server_name only if not yet set or set to default
+		if (tds_dstr_isempty(&connection->server_name) || strcmp(tds_dstr_cstr(&connection->server_name), TDS_DEF_SERVER) == 0)
 			res = tds_dstr_dup(&connection->server_name, &login->server_name);
 	}
 	if (login->tds_version)


freddy77:
Other people are using Azure without issue.
I would avoid, if possible, to change the server parsing, is already too complicated.
Looking at pymssql looks like you can pass a lot of parameters during connection so I would expect it possible to do it without any FreeTDS code change.

ThorstenEngel:
Dear Freddy, 

thank you for looking into this. Yes, I can understand that you don't want to touch this part of FreeTDS. I am not sure if you are aware that "Azure SQL" is a different product than the relatively new "Azure SQL Database managed instance" (my use case). 

I doublechecked everything and added PHP into my tests. Here is the result:

### **Without** my patch:
- Azure SQL (xy.database.windows.net) / pymssql: works with setting the username to "user@xy"
- Azure SQL (xy.database.windows.net) / PHP (PDO/dblib): works with setting the username to "user", but if adding ":1433" to the host, it needs "user@xy"
- Azure Managed SQL Server (xy.75a3c6410dc6.database.windows.net) / pymssql: does **not** work (patching pymssql is a viable alternative)
- Azure Managed SQL Server (xy.75a3c6410dc6.database.windows.net) / PHP (PDO/dblib): works with setting the username to "user", but with adding ":1433" it does not work anymore.

### **With** my patch:
- Azure SQL (xy.database.windows.net) / pymssql: works with setting the username to "user@xy" or to "user"
- Azure SQL (xy.database.windows.net) / PHP (PDO/dblib): works with setting the username to "user", or to "user@xy"
- Azure Managed SQL Server (xy.75a3c6410dc6.database.windows.net) / pymssql: works with setting the username to "user"
- Azure Managed SQL Server (xy.75a3c6410dc6.database.windows.net) / PHP (PDO/dblib): works with setting the username to "user"

### Final Options to allow pymssql to connect to Azure Managed SQL Server
a) make this (or a similar) change to FreeTDS
b) change pymssql to hand over the port directly in the login structure instead of appending it to the server_name.

I personally tend to change FreeTDS, because I think it is a bugfix. Changing pymssql will work in my case, but will not work if someone appends the port to the hostname. 

You decide, I can live with both results ;-) - and I am open to a further dialog.

Warm regards,

thorsten

ThorstenEngel:
addendum: fixing pymssql is not possible without a nonbreaking change in freetds. It would need something like:

#define DBSETLPORT(x,y)	dbsetlshort((x), (y), DBSETPORT)

to be able to set the **port** in a way that does **not use the server_name**.

moosakhalid:
Hi @ThorstenEngel ,
I am a newbie with pymssql and SQL server at  best but was facing exactly the issue that you described here. I'm trying to use pymssql to connect to an Azure Managed SQL instance and facing the same issue I believe. How do I use you fix? 
I have installed freetds and freetds-devel using yum on CentOS7, however you mentioned that connecting to Azure Managed SQL Instance needs openssl support , will I need to download and compile it(freetds)?

ThorstenEngel:
Hi @moosakhalid,
good to hear that you are not the only one. I had to learn the procedure as well. Yes, you have to download, apply my patch and compile FreeTDS by yourselves. It sounds harder than it is.

We are using Docker and that is why I created a Dockerfile which does exactly this. Looking at the DockerFile you can see all necessary steps. You even could use our container as I made it public.

Here is the Dockerfile: https://github.com/matrix-computer/python/blob/master/3.7/alpine/Dockerfile
Here ist the container: https://hub.docker.com/r/matrixcomputer/python 

I hope, this helps,

thorsten

freddy77:
DBSETLPORT could be an option. I'm currently considering however the parse_server_name_for_port to be buggy in the sense that should probably change login->server_name and not connection->server_name

freddy77:
What about this patch?
```diff
diff --git a/src/tds/config.c b/src/tds/config.c
index 39959cbf..101ecdab 100644
--- a/src/tds/config.c
+++ b/src/tds/config.c
@@ -187,6 +187,10 @@ tds_read_config_info(TDSSOCKET * tds, TDSLOGIN * login, TDSLOCALE * locale)
                                }
                                found = true;
                        }
+                       if (!tds_dstr_dup(&login->server_name, &connection->server_name)) {
+                               tds_free_login(connection);
+                               return NULL;
+                       }
                }
        }
        if (!found) {
```

ThorstenEngel:
Dear Freddy,

this fix worked in my test cases (managed sql Server on Azure & on premise)! Great!

Regards,

Thorsten

Von meinem iPhone gesendet

Am 09.05.2019 um 22:45 schrieb freddy77 <notifications@github.com<mailto:notifications@github.com>>:


What about this patch?

diff --git a/src/tds/config.c b/src/tds/config.c
index 39959cbf..101ecdab 100644
--- a/src/tds/config.c
+++ b/src/tds/config.c
@@ -187,6 +187,10 @@ tds_read_config_info(TDSSOCKET * tds, TDSLOGIN * login, TDSLOCALE * locale)
                                }
                                found = true;
                        }
+                       if (!tds_dstr_dup(&login->server_name, &connection->server_name)) {
+                               tds_free_login(connection);
+                               return NULL;
+                       }
                }
        }
        if (!found) {

freddy77:
Good, merged

PR #281, opened 2019-04-09
fixes for tds_listen in server/login.c

maharmstone:
Here's some fixes for tds_listen in server/login.c.

The first changes AF_INET to AF_INET6, allowing it to accept connexions via IPv6. It'll still work with IPv4.

The second updates the state to TDS_IDLE when a client connects. Without this, it'll remain as TDS_DEAD, meaning that the call to tds_alloc_read_login in server/unittest.c will always fail.

### DIFF
--- a/src/server/login.c
+++ b/src/server/login.c
@@ -75,15 +75,15 @@ TDSSOCKET *
 tds_listen(TDSCONTEXT * ctx, int ip_port)
 {
 	TDSSOCKET *tds;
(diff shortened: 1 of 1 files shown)


freddy77:
Merged the IDLE one. Not sure about IPv6 portability, can I add some conditional compilation?

maharmstone:
Cheers. Feel free...

freddy77:
Done and merged, thanks

PR #304, opened 2019-08-30
New CS_DATABASE property

### LINKED ISSUE #303: CTLib unusable with Microsoft Azure
Similar to #160, the C library in its current form cannot be used with Microsoft Azure databases. By default CTLib will connect to the "master" database, and attempts to switch to another database yield the error:

> USE statement is not supported to switch between databases. Use a new connection to connect to a different database.

I have worked around the issue by defining a new `CS_DATABASE` property that calls `tds_dstr_copy(&tds_login->database, set_buffer)` inside of `ct_con_props`, so that the database can be defined before the initial connection. I can contribute this code if desired.

evanmiller:
Microsoft Azure does not allow switching databases after the connection has been made, so we need a way to define the database name prior to the call to `ct_connect`. Introduce a new, optional `CS_DATABASE` property that can be configured with a string value using `ct_con_props`.

I have defined the constant / enum as 9302, as it appears that the other 93XX properties (`CS_PORT` and `CS_CLIENTCHARSET`) are also inventions of FreeTDS.

Fixes #303


freddy77:
Merged, added GET support

PR #318, opened 2019-12-07
Fixup/flt8 to decimal convert

detule:
Hi @freddy77 

Came across a failure in trying to write a FLOAT to a DECIMAL(p, s) column in SQL SERVER.  I believe I tracked it down to the use of `sprintf(tmp_str, "%.16g", the_value); ` to write the float (source/type) to a string (which is subsequently parsed into a DECIMAL).

In particular, the `%g` conversion specifier seems a little suspect - in my case it was converting my value `0.00001` to `1e-05`, which is not something that `string_to_numeric` is equipped to handle.

Look forward to your feedback.

### DIFF
--- a/src/tds/convert.c
+++ b/src/tds/convert.c
@@ -1692,7 +1692,7 @@ tds_convert_flt8(const TDS_FLOAT* src, int desttype, CONV_RESULT * cr)
 		break;
 	case SYBNUMERIC:
 	case SYBDECIMAL:
-		sprintf(tmp_str, "%.16g", the_value);
+		sprintf(tmp_str, "%.*f", cr->n.scale, the_value);
 		return stringz_to_numeric(tmp_str, cr);
 		break;
 		/* not allowed */
--- a/src/tds/unittests/convert.c
+++ b/src/tds/unittests/convert.c
@@ -333,7 +333,7 @@ main(int argc, char **argv)
 				srctype, tds_prtype(srctype), srclen,
 				desttype, tds_prtype(desttype));
 
-			if (result == TDS_CONVERT_NOAVAIL)
+			if (result != TDS_CONVERT_NOAVAIL)
 				exit(1);
 		}
 
--- a/src/tds/unittests/convert.c
+++ b/src/tds/unittests/convert.c
@@ -242,7 +242,7 @@ main(int argc, char **argv)
 			srclen = sizeof(tds_uint8);
 			break;
 		case SYBFLT8:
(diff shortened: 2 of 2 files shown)


src/tds/convert.c
```diff
@@ -1692,7 +1692,7 @@ tds_convert_flt8(const TDS_FLOAT* src, int desttype, CONV_RESULT * cr)
 		break;
 	case SYBNUMERIC:
 	case SYBDECIMAL:
-		sprintf(tmp_str, "%.16g", the_value);
+		sprintf(tmp_str, "%.*f", cr->n.scale, the_value);
```
freddy77: I would say yes! Also it's the same code in tds_convert_real.

src/tds/unittests/convert.c
```diff
@@ -333,7 +333,7 @@ main(int argc, char **argv)
 				srctype, tds_prtype(srctype), srclen,
 				desttype, tds_prtype(desttype));
 
-			if (result == TDS_CONVERT_NOAVAIL)
+			if (result != TDS_CONVERT_NOAVAIL)
```
freddy77: no, this test is here to make it sure that we support this conversion, as tds_willconvert told us.
not said however that maybe the test could be more strict.

src/tds/unittests/convert.c
```diff
@@ -242,7 +242,7 @@ main(int argc, char **argv)
 			srclen = sizeof(tds_uint8);
 			break;
 		case SYBFLT8:
-			tds_float = 3.14159;
+			tds_float = 0.000001;
```
freddy77: not sure if would be better for this case to extend t0007 test instead. I didn't manage to reproduce the error under Linux. Which environment are you using?

src/tds/unittests/convert.c
detule: I was a bit surprised to find the test not fail when I expected it to.  
Thanks for the explanation that it's not restrictive by design, so I'll take out the patch.

(For what is worth, the test passes for me as such / strict).

src/tds/unittests/convert.c
detule: Looks fine, nice work - reflects the problem as well.

detule:
Thanks @freddy77 Will:

* Take out the unittesting changes.
* Update the commit message on the content patch to reflect
  * prefix convention
  * more robust explanation of patch

detule:
Abandoned in favor of https://github.com/FreeTDS/freetds/pull/319

PR #319, opened 2019-12-09
tds: More robust conversion from FLT to DECIMAL/NUMERIC

detule:
Conversion follows flt->string->numeric code-path.  In the first step
(flt->string via `sprintf`) we use `%f` conversion specifier to prevent
receiving a string with scientific notation back (as is the case,
potentially, with `%g`).


freddy77:
Merge and ported to 1.1. Thanks

PR #329, opened 2020-04-14
ignore computed columns when doing bulk copy

maharmstone:
This patch fixes freetds so that it ignores computed columns when doing a bulk insert, in the same manner as with TIMESTAMP columns. Without it, tds_bcp_start_insert_stmt will fail, even if the computed column hasn't been directly specified.

The flag constant for computed columns, 0x20, comes from https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/58880b9f-381c-43b2-bf8b-0727a98c4f4c.


freddy77:
Merged, copied message here in the commit message, it was more clear

PR #339, opened 2020-05-10
tds: Fix for TDS7+ server login and queries

smorey2:
TDS connections from Microsoft products require TDS72+, and SSMS requires product_version. Previously all TDS71+ became TDS71.
TDS72+ query adds a query head.

Signed-off-by: Sky Morey <smorey@degdigital.com>

Server connection from .net sql connector, or SSMS studio, required a newer TDS.
Open to reworking or regrouping these changes as needed.

### DIFF
--- a/src/server/query.c
+++ b/src/server/query.c
@@ -78,14 +78,14 @@ tds_get_query_head(TDSSOCKET * tds, TDSHEADERS * head)
 
 	if (IS_TDS72_PLUS(tds->conn)) {
 		int qn_len = 0;
-		const char *qn_msgtext = NULL;
-		const char *qn_options = NULL;
+		char *qn_msgtext = NULL;
+		char *qn_options = NULL;
 		size_t qn_msgtext_len = 0;
 		size_t qn_options_len = 0;
 
 		qn_len = tds_get_int(tds) - 4 - 18;             /* total length */
-		tds_get_int(tds, 18);                          /* length: transaction descriptor, ignored */
-		tds_get_smallint(tds, 2);                      /* type: transaction descriptor, ignored */
+		tds_get_int(tds);                          /* length: transaction descriptor, ignored */
+		tds_get_smallint(tds);                      /* type: transaction descriptor, ignored */
 		tds_get_n(tds, tds->conn->tds72_transaction, 8);  /* transaction */
 		tds_get_int(tds, 1);                           /* request count, ignored */
 		if (qn_len != 0) {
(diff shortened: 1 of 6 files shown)


src/server/query.c
```diff
@@ -65,17 +65,71 @@ tds_lastpacket(TDSSOCKET * tds)
...
+ * get query packet of a given type
+ * \tds
+ * \param head         extra information to put in a TDS7 header
+ */
+static TDSRET
+tds_get_query_head(TDSSOCKET * tds, TDSHEADERS * head)
+{
+	int more;
+
+	if (IS_TDS72_PLUS(tds->conn)) {
```
freddy77, line 79: maybe return here to reduce function indentation?

freddy77:
Are you using some experimental server? Or is this affecting the pool server?

src/server/query.c
smorey2: Sure can do, was just following existing code from the query side. I prefer exit-early. but also wanted to follow existing code format.

https://github.com/FreeTDS/freetds/blob/master/src/tds/query.c # tds_start_query_head

smorey2:
Not using the pool server, attempting to utilize the hosting parts of the server folder.

I have numerous Windows Services, and have built a .net wrapper around FreeTDS. Will host a TDS socket, and connect to it with .net, and SQL Server Management Studio.

Another project I had considered, was a Windows Service with a TDS endpoint to a Sqlite database.

recent .net client libraries no longer accept anything before TDS72.

smorey2:
I had spend some time tracing both sides of code. Anything higher than TDS71 became 71, so some way of providing that. I had initially placed on read_login, along with the product_version.
not used: `tds_alloc_read_login(TDSSOCKET * tds, TDS_USMALLINT tds_version, TDS_UINT product_version);`

but login_ack had the progname (product_name) instead of using conn->product_name.
existing: `void tds_send_login_ack(TDSSOCKET * tds, const char *progname);`

I was confused because login_ack uses a product_name argument, ignoring the conn->product_name, but uses the conn->product_version.

so thought this is where product_version should be. this does however change pool.
new: `void tds_send_login_ack(TDSSOCKET * tds, const char *progname, TDS_UINT product_version);`

if login_ack uses arg progname (rename to product_name?), and arg product_version. should they only be args, or set on conn (currently setting version on conn)?

None of these interface changes are required since I can manually set tds_version, and product_version, on conn before login_ack. This did take me a while to figure out, and I thought it could be more beneficial for others to have it explicit.

smorey2:
closing, will create an issue.

PR #343, opened 2020-05-21
tdssrv: support TDS72+ logins

smorey2:
This fixes FreeTDS#340.

Signed-off-by: Sky Morey <smorey@degdigital.com>

These where the simple changes I had discussed, had committed days ago. but have been busy.

Also not sure where to property send pull requests, master or branch

### DIFF
--- a/src/server/query.c
+++ b/src/server/query.c
@@ -29,6 +29,7 @@
 
 static char *query;
 static size_t query_buflen = 0;
+static TDSHEADERS head;
 
 /**
  * Read a TDS5 tokenized query.
@@ -63,6 +64,58 @@ tds_lastpacket(TDSSOCKET * tds)
 		return 1;
 	
 	return tds->in_buf[1] != 0;
+}
+
+/**
+ * get query packet of a given type
+ * \tds
+ * \param head         extra information to put in a TDS7 header
+ */
+static TDSRET
+tds_get_query_head(TDSSOCKET * tds, TDSHEADERS * head)
+{
+	if (!IS_TDS72_PLUS(tds->conn)) {
+		return TDS_SUCCESS;
+	}
+
+	int qn_len = 0, more;
+	char *qn_msgtext = NULL;
+	char *qn_options = NULL;
+	size_t qn_msgtext_len = 0;
+	size_t qn_options_len = 0;
+
+	qn_len = tds_get_int(tds) - 4 - 18;             /* total length */
+	tds_get_int(tds);                          /* length: transaction descriptor, ignored */
+	tds_get_smallint(tds);                      /* type: transaction descriptor, ignored */
+	tds_get_n(tds, tds->conn->tds72_transaction, 8);  /* transaction */
+	tds_get_int(tds);                           /* request count, ignored */
+	if (qn_len != 0) {
+		if (!head) {
+			tds_set_state(tds, TDS_IDLE);
+			return TDS_FAIL;
+		}
+
+		qn_len = tds_get_int(tds);                   /* length: query notification */
+		tds_get_smallint(tds);                      /* type: query notification, ignored */
+		qn_msgtext_len = tds_get_smallint(tds);  /* notifyid */
+		if (qn_msgtext_len > 0) {
+			qn_msgtext = (char *) realloc(qn_msgtext, qn_msgtext_len);
+			tds_get_n(tds, qn_msgtext, qn_msgtext_len);
+		}
+
+		qn_options_len = tds_get_smallint(tds);  /* ssbdeployment */
+		if (qn_options_len > 0) {
+			qn_options = (char *) realloc(qn_options, qn_options_len);
+			tds_get_n(tds, qn_options, qn_options_len);
+		}
+		more = tds->in_len - tds->in_pos;
+		if (more)
+			head->qn_timeout = tds_get_int(tds);        /* timeout */
+
+		head->qn_options = qn_options;
+		head->qn_msgtext = qn_msgtext;
+	}
+	return TDS_SUCCESS;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


freddy77:
Branches are for stable version, opened mainly for bug fixing or small improvements not causing regressions. Main development on master. So for these changes master.

It looks like your changes do not compile correctly, see Appveyor results (unfortunately there are also some system failures, but last build is a real one).

src/server/query.c
```diff
@@ -65,6 +66,58 @@ tds_lastpacket(TDSSOCKET * tds)
...
+ * \param head         extra information to put in a TDS7 header
+ */
+static TDSRET
+tds_get_query_head(TDSSOCKET * tds, TDSHEADERS * head)
+{
+	if (!IS_TDS72_PLUS(tds->conn)) {
+		return TDS_SUCCESS;
+	}
+
+	int qn_len = 0, more;
```
freddy77, line 81: we support very old systems that does not like to mix declaration and statements, declarations should be at the beginning of blocks.

src/server/unittest.c
```diff
@@ -65,6 +65,8 @@ main(int argc, char **argv)
 		exit(1);
 	}
 	dump_login(login);
+	// tds->conn->tds_version = 0x702;
+	// tds->conn->product_version = G.TDS_MS_VER(10, 0, 6000);
```
freddy77, line 69: also some systems not liking `//` style comments.

src/server/query.c
```diff
@@ -65,6 +66,58 @@ tds_lastpacket(TDSSOCKET * tds)
...
+	size_t qn_msgtext_len = 0;
+	size_t qn_options_len = 0;
+
+	qn_len = tds_get_int(tds) - 4 - 18;             /* total length */
+	tds_get_int(tds);                          /* length: transaction descriptor, ignored */
+	tds_get_smallint(tds);                      /* type: transaction descriptor, ignored */
+	tds_get_n(tds, tds->conn->tds72_transaction, 8);  /* transaction */
+	tds_get_int(tds);                           /* request count, ignored */
+	if (qn_len != 0) {
+		if (!head) {
```
freddy77, line 93: would it not better to require `head` not to be NULL calling this function?

src/server/query.c
```diff
@@ -65,6 +66,58 @@ tds_lastpacket(TDSSOCKET * tds)
...
+		tds_get_smallint(tds);                      /* type: query notification, ignored */
+		qn_msgtext_len = tds_get_smallint(tds);  /* notifyid */
+		if (qn_msgtext_len > 0) {
+			qn_msgtext = (char *) realloc(qn_msgtext, qn_msgtext_len);
+			tds_get_n(tds, qn_msgtext, qn_msgtext_len);
+		}
+
+		qn_options_len = tds_get_smallint(tds);  /* ssbdeployment */
+		if (qn_options_len > 0) {
+			qn_options = (char *) realloc(qn_options, qn_options_len);
```
freddy77, line 108: This code will crash if `realloc` returns NULL. Other part is handling these conditions. Maybe for the server just using a kind of `xrealloc` would be enough.

src/server/unittest.c
smorey2: changed to /* */ format

src/server/query.c
smorey2: removed test, head required

src/server/query.c
smorey2: moved to after declarations

src/server/query.c
smorey2: was mimic code from tds/query.c
tds_start_query_head(TDSSOCKET *tds, unsigned char packet_type, TDSHEADERS * head)

where was just using realloc. all realloc would have the same issue.

smorey2:
should I be building unittests also, maybe can use tds/unittest/t0001.c-t0009.c as test clients. not sure how to properly process launch. Or if I should just start a thread, and include.

src/server/query.c
freddy77: It's a bit difference between checking for NULL handling that case returning error and ignoring and use as it were no NULL causing a crash.

src/server/unittest.c
freddy77: I cannot see the change, maybe you forgot to push it upstream?

src/server/query.c
smorey2: I switch all out with xrealloc. I wanted to try and build some unittests.

Also noticed a few things that would need to be corrected. But probably better incrementally.

Thank you for working with me freddy77. better then being in a vacuum.

smorey2:
./.libs/libtdssrv.a(query.o): In function `tds_get_query':
.../server/query.c:52: undefined reference to `xrealloc'
./.libs/libtdssrv.a(query.o): In function `tds_get_generic_query':
.../server/query.c:224: undefined reference to `xrealloc'
.../server/query.c:157: undefined reference to `xrealloc'
./.libs/libtdssrv.a(query.o): In function `tds_get_query_head':
.../server/query.c:97: undefined reference to `xrealloc'
.../server/query.c:103: undefined reference to `xrealloc'
.../server/query.c:174: more undefined references to `xrealloc' follow

freddy77:
If you agree and Appveyor pass I would squash the commits and merge the result.

About dependencies for Ubuntu one option is to check from https://packages.ubuntu.com the `freetds` package (https://packages.ubuntu.com/eoan/freetds-bin) the `.dsc` file (right column), if you open you will see a line like

```
Build-Depends: debhelper (>= 10), autotools-dev, unixodbc-dev (>= 2.2.11-10), libglib2.0-dev | libglib-dev, libgnutls28-dev, libreadline-dev, libkrb5-dev, pkg-config
```
(not sure why Ubuntu needs glib, maybe they added some utilities).

For the unittests no idea, maybe some base tests in src/tds/unittests can be used but probably the most complicated not. Maybe a specific script that launch only some of them.

smorey2:
everything looks good. And I will look at a couple unittests, following t0001.c. Might need help and review with the build pipeline. Thanks.

freddy77:
Merged, thanks

PR #344, opened 2020-06-01
odbc: SQLColumns: Ability to operate out-of-catalog

detule:
Hi @freddy77 

Currently I am unable to execute an out-of-current-catalog call to `SQLColumns`.  To highlight this in an example, I modified one of the unit-tests [here](https://github.com/detule/freetds/commit/34ec69e7e07d28f40c962e21b435ad01c72205d2) (unit test mod not submitted here as part of the PR).

With this  patch, we call `catalog.sp_columns` (catalog sproc, rather than system sproc), whenever a non-trivial catalog is supplied.

* I am not sure how sybase would be affected - happy to guard the added code inside an `if MSSQL` block if you think that's appropriate.
* Also happy to submit this mod or another unit test for this functionality.

Look forward to getting your feedback.

### DIFF
--- a/src/odbc/odbc.c
+++ b/src/odbc/odbc.c
@@ -4785,11 +4785,27 @@ ODBC_FUNC(SQLColumns, (P(SQLHSTMT,hstmt), PCHARIN(CatalogName,SQLSMALLINT),	/* o
 	WIDE))
 {
 	int retcode;
+	DSTR catalog_name = DSTR_INITIALIZER;
+	SQLSMALLINT catalog_size;
+	const char *proc = NULL;
 
 	ODBC_ENTER_HSTMT;
 
+	if (cbCatalogName == SQL_NTS) {
+		if (!odbc_dstr_copy(stmt->dbc, &catalog_name, cbCatalogName, szCatalogName))
+			goto memory_error;
+		catalog_size = tds_dstr_len(&catalog_name);
+	} else {
+		catalog_size = cbCatalogName;
+	}
+
+	proc = "sp_columns";
+	if (catalog_size > 0) {
+		proc = "..sp_columns";
+	}
+
 	retcode =
-		odbc_stat_execute(stmt _wide, "sp_columns", TDS_IS_MSSQL(stmt->dbc->tds_socket) ? 5 : 4,
+		odbc_stat_execute(stmt _wide, proc, TDS_IS_MSSQL(stmt->dbc->tds_socket) ? 5 : 4,
 				  "P@table_name", szTableName, cbTableName, "P@table_owner", szSchemaName,
 				  cbSchemaName, "O@table_qualifier", szCatalogName, cbCatalogName, "P@column_name", szColumnName,
 				  cbColumnName, "V@ODBCVer", (char*) NULL, 0);
@@ -4803,6 +4819,13 @@ ODBC_FUNC(SQLColumns, (P(SQLHSTMT,hstmt), PCHARIN(CatalogName,SQLSMALLINT),	/* o
 		if (TDS_IS_SYBASE(stmt->dbc->tds_socket))
 			stmt->special_row = ODBC_SPECIAL_COLUMNS;
 	}
+
+	tds_dstr_free(&catalog_name);
+	ODBC_EXIT_(stmt);
+
+memory_error:
+	tds_dstr_free(&catalog_name);
+	odbc_errs_add(&stmt->errs, "HY001", NULL);
 	ODBC_EXIT_(stmt);
 }
 
--- a/src/odbc/odbc.c
+++ b/src/odbc/odbc.c
@@ -4785,24 +4785,12 @@ ODBC_FUNC(SQLColumns, (P(SQLHSTMT,hstmt), PCHARIN(CatalogName,SQLSMALLINT),	/* o
 	WIDE))
 {
 	int retcode;
-	DSTR catalog_name = DSTR_INITIALIZER;
-	SQLSMALLINT catalog_size;
-	const char *proc = NULL;
+	const char *proc = "sp_columns";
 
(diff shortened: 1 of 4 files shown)


src/odbc/odbc.c
```diff
@@ -4785,11 +4785,27 @@ ODBC_FUNC(SQLColumns, (P(SQLHSTMT,hstmt), PCHARIN(CatalogName,SQLSMALLINT),	/* o
...
+	}
+
+	proc = "sp_columns";
+	if (catalog_size > 0) {
+		proc = "..sp_columns";
+	}
+
 	retcode =
-		odbc_stat_execute(stmt _wide, "sp_columns", TDS_IS_MSSQL(stmt->dbc->tds_socket) ? 5 : 4,
+		odbc_stat_execute(stmt _wide, proc, TDS_IS_MSSQL(stmt->dbc->tds_socket) ? 5 : 4,
```
freddy77, line 4808: I have the feeling that just replacing `"sp_columns"` with `"..sp_columns"` here will solve the problem. Please send the test case, I can integrate in some way and use to test all various server types and versions.

detule: Added to the PR; For the test, I modified the only one where I saw `SQLColumns` being tested - `stats` - it's pretty ugly, sorry.

Thanks for considering it.

src/odbc/unittests/stats.c
```diff
@@ -49,13 +48,25 @@ TestProc(const char *type, const char *expected)
...
 {
 	char sql[256];
+	char *schema = NULL;
+
+	if (catalog) {
+		schema = "dbo";
+		sprintf(sql, "DROP DATABASE IF EXISTS %s", catalog);
+		odbc_command(sql);
+		sprintf(sql, "CREATE DATABASE %s", catalog);
+		odbc_command(sql);
```
freddy77, line 61: requirements for the user are not so high to be able to create and destroy databases, see http://www.freetds.org/userguide/confirminstall.htm#TESTS, but well, nothing is set on stone. But I would instead asking to create (manually in the documentation) a database and have the user as owner of this database.

detule: Sure, I'll:

* Create the `testDB` DB in misc/appveyor.xml via `sqlcmd`
* Update the requirements in doc/userguide.xml
* Simplify the unit-test here knowing that the DB exists.

freddy77: Or would be possible, knowing `tempdb` and `master` are always available to save the current database name, move to either `tempdb` or `master` and use original database for tests. Not sure how easy but should be possible. Won't work on Azure as I think there you need to specify a database at login time and you can't switch but it's not a big deal. Or require to create a `freetds_test` database (or simething similar), create it on Appveyor and document on the user guide.

detule: The `freetds_test` DB seemed relatively simple - added a new commit.

Thanks again

detule:
Hi @freddy77 gentle ping.  Let me know if I can do anything else here.

Thanks

freddy77:
> Hi @freddy77 gentle ping. Let me know if I can do anything else here.
> 
> Thanks

Hi, sorry, a bit busy with the release and other stuff. The most concern part is the additional requirement for testing. Updating now anyway some script I have, I still needs to test it under Sybase.

freddy77:
Got some updates. Mostly done with additional setup for tests however Sybase 15 does not like much the change, I'll have to understand why (unless you are able to check this too).

src/odbc/odbc.c
```diff
@@ -4785,11 +4785,27 @@ ODBC_FUNC(SQLColumns, (P(SQLHSTMT,hstmt), PCHARIN(CatalogName,SQLSMALLINT),	/* o
 	WIDE))
 {
 	int retcode;
+	DSTR catalog_name = DSTR_INITIALIZER;
+	SQLSMALLINT catalog_size;
+	const char *proc = NULL;
 
 	ODBC_ENTER_HSTMT;
 
+	if (cbCatalogName == SQL_NTS) {
```
freddy77, line 4794: reading again this if caller specifies a value > 0 (that is a valid size like `"test"` with len 4) catalog won't be prepended. Why not just remove the if and copy always?

src/odbc/odbc.c
```diff
@@ -4785,11 +4785,27 @@ ODBC_FUNC(SQLColumns, (P(SQLHSTMT,hstmt), PCHARIN(CatalogName,SQLSMALLINT),	/* o
 	WIDE))
 {
 	int retcode;
+	DSTR catalog_name = DSTR_INITIALIZER;
+	SQLSMALLINT catalog_size;
+	const char *proc = NULL;
```
freddy77, line 4790: Why not initializing to `"sp_columns"` directly here?

freddy77:
So, the issue with Sybase is that Sybase does not like quoting in RPC calls. Unfortunately you can create databases with names which contains quotes, spaces or dots so I'm not sure how to correctly quote them. However I'm wondering why this was not detected before (potentially not tested correctly!)

src/odbc/odbc.c
detule: Because I didn't know `odbc_get_string_size` existed :) Will change, thanks for the tip.

src/odbc/odbc.c
detule: Good feedback; will change, thanks.

src/odbc/odbc.c
detule: Done, let me know if I missed anything.

src/odbc/odbc.c
detule: Done, let me know if I missed anything.

freddy77:
Merged, thanks

PR #346, opened 2020-06-03
another VMS build update

craigberry:


freddy77:
Merged in `Branch-1_2`, beside the last, see comment

src/server/login.c
```diff
@@ -77,7 +77,7 @@ tds_listen(TDSCONTEXT * ctx, int ip_port)
 	TDSSOCKET *tds;
 	TDS_SYS_SOCKET fd, s;
 	socklen_t len;
-#ifdef AF_INET6
+#if defined(AF_INET6) && !defined(__VMS) /* in6addr_any may or may not exist on VMS */
```
freddy77, line 80: very similar code in `src/tds/net.c`, I suspect that there we have the right includes, I think would be better to add required includesin `src/server/login.c` too.

craigberry: net.c is not that similar.  Specifically it does not reference the global variable in6addr_any.  This variable is global to the system, not just the program, and is generally defined in libc somewhere.  On VMS, there is more than one network stack and it depends on which one you are running whether the variable has a value defined (it's in headers regardless); if you don't have it, you get linker errors.  There is another way to do it by defining your own constant and initializing to IN6ADDR_ANY_INIT at compile time.  I will see if I can get that working.  The [RFC](https://tools.ietf.org/html/rfc2553#section-3.8) has more details.

freddy77: Should not be all zeroes? Maybe a simple `memset` would solve the problem?

craigberry: The memset would probably work.  I think the documented way would look more like this:

```
#if defined(AF_INET6)
	struct sockaddr_in6 sin;
	struct in6_addr anyaddr = IN6ADDR_ANY_INIT;

	sin.sin6_addr = anyaddr;
	sin.sin6_port = htons((short) ip_port);
	sin.sin6_family = AF_INET6;

	s = socket(AF_INET6, SOCK_STREAM, 0);
#else

```

I haven't tried this yet and may not get to it until tomorrow.

freddy77: No rush. In the meantime the `memset` works perfectly on Linux.

craigberry: Sorry for the slow reply.  I just tried RC2 and the memset works fine on VMS.

PR #348, opened 2020-06-15
string_to_datetime: reject garbled strings

maharmstone:
There's an apparent bug in `string_to_datetime`, where it only checks if a string is garbled within the loop, not after it. This results in it incorrectly converting invalid strings of only one token, i.e. without a comma, to 1900-01-01 00:00:00.

I'm fairly sure this isn't intentional behaviour, as it doesn't tally with what MSSQL does - it throws error 241 if you try to do, e.g., `SELECT CONVERT(DATETIME, 'hello');`.

### DIFF
--- a/src/tds/convert.c
+++ b/src/tds/convert.c
@@ -2211,6 +2211,13 @@ string_to_datetime(const char *instr, TDS_UINT len, int desttype, CONV_RESULT * 
 		tok = strtok_r(NULL, " ,", &lasts);
 	}
 
+	if (current_state == STRING_GARBLED) {
+		tdsdump_log(TDS_DBG_INFO1,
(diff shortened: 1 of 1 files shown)


freddy77:
What about something like https://github.com/freddy77/freetds/commit/7d14e41205ab1824a0897f3334e14386ae3492e2 ?

maharmstone:
Looks good to me.

freddy77:
@maharmstone  Updated with better commit message, I included your  name. How about it ?

https://github.com/FreeTDS/freetds/commit/f29eb8719bf1071f7dbf386e3153273dad2a72f4

freddy77:
@maharmstone, any comment?

maharmstone:
Looks great, thank you!

PR #353, opened 2020-08-10
Added ODBC timeout parameter for odbc.ini and connection string

ialden:


src/odbc/connectparams.c
```diff
@@ -844,6 +844,9 @@ ODBCINSTGetProperties(HODBCINSTPROPERTY hLastProperty)
 	hLastProperty = definePropertyList(hLastProperty, odbc_param_Encryption, TDS_STR_ENCRYPTION_OFF, aEncryption, sizeof(aEncryption),
 		"The encryption method.");
 
+	hLastProperty = definePropertyString(hLastProperty, odbc_param_Timeout, "10", 
+		"Connection timeout.");
+
```
freddy77, line 849: The code change is good but the commit message could be misleading `Added default 10 seconds for timeout` does not tell that this change affects the unixODBC configuration only.

ialden: Ok I have amended the commit message.

freddy77:
Merged, thanks

PR #354, opened 2020-08-20
Check if memory is allocated before use it

klebertarcisio:
Check if memory is allocated before use it.

### DIFF
--- a/src/apps/datacopy.c
+++ b/src/apps/datacopy.c
@@ -655,7 +655,10 @@ transfer_data(const BCPPARAMDATA * params, DBPROCESS * dbsrc, DBPROCESS * dbdest
 	}
 
 	srcdata = (MIGCOLDATA *) calloc(sizeof(MIGCOLDATA), src_numcols);
-
+	if(srcdata == NULL){
+		printf("Could not allocate memory to srcdata\n");
+		return FALSE;
+	}
 	for (col = 0; col < src_numcols; col++) {
 
 		/* Find out if there is an identity column. */
--- a/src/ctlib/ct.c
+++ b/src/ctlib/ct.c
@@ -3372,7 +3372,10 @@ ct_param(CS_COMMAND * cmd, CS_DATAFMT * datafmt, CS_VOID * data, CS_INT datalen,
 		}
 
 		param = tds_new0(CSREMOTE_PROC_PARAM, 1);
-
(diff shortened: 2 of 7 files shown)


freddy77:
Thanks for taking care of the missing unchecked allocations

src/apps/datacopy.c
```diff
@@ -655,7 +655,10 @@ transfer_data(const BCPPARAMDATA * params, DBPROCESS * dbsrc, DBPROCESS * dbdest
 	}
 
 	srcdata = (MIGCOLDATA *) calloc(sizeof(MIGCOLDATA), src_numcols);
-
+	if(srcdata == NULL){
```
freddy77, line 658: About syntax, code uses a space after if and before bracket, also prefer `!ptr` instead of `ptr == NULL`, so

```c
if (!srcdata) {
```

I can do the change if you prefer

src/apps/datacopy.c
```diff
@@ -655,7 +655,10 @@ transfer_data(const BCPPARAMDATA * params, DBPROCESS * dbsrc, DBPROCESS * dbdest
...
+		printf("Could not allocate memory to srcdata\n");
```
freddy77, line 659: I would use `stderr` for errors.
Is it not `Could not allocate memory FOR srcdata` ?

src/ctlib/ct.c
```diff
@@ -3372,7 +3372,10 @@ ct_param(CS_COMMAND * cmd, CS_DATAFMT * datafmt, CS_VOID * data, CS_INT datalen,
 		}
 
 		param = tds_new0(CSREMOTE_PROC_PARAM, 1);
-
+		if(!param){
+			printf("Could not allocate memory to param\n");
```
freddy77, line 3376: libraries should not output on standard error, each library (in this case CT-Library) has a way to report errors.

src/pool/main.c
```diff
@@ -114,6 +114,10 @@ pool_init(const char *name, const char *config_path)
 	/* initialize the pool */
 
 	pool = tds_new0(TDS_POOL, 1);
+	if(!poll){
```
freddy77, line 117: Typo: `poll` -> `pool`

klebertarcisio:
Hi @freddy77 . I did some changes. Tell me if you need anything more.

PR #358, opened 2020-09-08
Fix compilation error

klebertarcisio:
I think this should be a comment.

Regards


freddy77:
Sure. That possibly means that `USING_SYBEBCNN` is never defined.

freddy77:
Merged

klebertarcisio:
Thanks @freddy77

PR #375, opened 2020-12-16
Fix datacopy logging

darktemplarbasealt:
Use stderr for logging when it's enabled instead of discarding logs.


freddy77:
Merged, thanks

PR #383, opened 2021-02-24
Avoiding memory leak and null pointer dereference

klebertosantos:
This pull request aims to fix some memory leak and null pointer dereference.

[Memory Leak](https://cwe.mitre.org/data/definitions/401.html)
[Null pointer dereference](https://cwe.mitre.org/data/definitions/476.html)

### DIFF
--- a/src/apps/bsqldb.c
+++ b/src/apps/bsqldb.c
@@ -523,7 +523,7 @@ print_results(DBPROCESS *dbproc)
 
 				bynames = strdup("by (");
 				if(!bynames) {
-					printf('no memory to allocate');
+					printf("no memory to allocate");
 					exit(1);
 				}
 				for (iby=0; iby < nby; iby++) {
--- a/src/apps/bsqldb.c
+++ b/src/apps/bsqldb.c
@@ -522,6 +522,10 @@ print_results(DBPROCESS *dbproc)
 				bylist = dbbylist(dbproc, c+1, &nby);
 
 				bynames = strdup("by (");
+				if(!bynames) {
+					printf('no memory to allocate');
+					exit(1);
+				}
 				for (iby=0; iby < nby; iby++) {
 					char *s = NULL; 
 					int ret = asprintf(&s, "%s%s%s", bynames, dbcolname(dbproc, bylist[iby]), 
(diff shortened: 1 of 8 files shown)


fziglio:
At least it would be good if the code compiled.

klebertarcisio:
Hi @fziglio, all pull requests have a compilation error, would it be a problem with the configuration of continuous integration on github?

klebertosantos:
Hi @fziglio, all pull requests have a compilation error, would it be a problem with the configuration of continuous integration on github?

fziglio:
@klebertosantos, the issue is that your code is not even valid C code, not the CI.

klebertosantos:
You are right! @fziglio  Now I made the correction by placing double quotes instead of single quotes. Thanks!

freddy77:
Ok, better. Style, space after if. Error goes to error, not normal output. Library should not print on either standard or error output but return proper error.

klebertosantos:
**Hi, @freddy77 Thanks for the feedback.** 
>Ok, better. Style, space after if.

Perfect, I can do it.

>Library should not print on either standard or error output but return proper error.

Could you tell me in which places should I not have the print?

>In a lot of other places the message is simply "out of memory".

Ok, I can replace "no memory to allocate" to "out of memory".

fziglio:
> > Library should not print on either standard or error output but return proper error.
> 
> Could you tell me in which places should I not have the print?
> 
Sure,  `src/ctlib/ct.c`,  `src/dblib/dbpivot.c` and `src/tds/data.c`.

 On another subject, `src/tds/data.c` has a label you can probably reuse.

klebertosantos:
Hi @fziglio, thanks for the feedback. I did some changes!

freddy77:
About error on standard error I would replace `printf("out of memory");` to `fprintf(stderr, "out of memory\n");`

About `src/tds/data.c` I would add a label `error_memory` at the same line of `error_type` and jump to it instead. There's also another `tds_new0` in the same `tds_variant_get` function.

Could you squash the commits together? They are mainly fixes to other patches.

freddy77:
@klebertosantos what about this commit https://github.com/FreeTDS/freetds/commit/b7483a5740f89fcbcb8cd8fa587132a8eaf84509 ? I updated a bit your ones

klebertarcisio:
Hi @freddy77, sorry for the delay, i will update very soon.

fziglio:
> Hi @freddy77, sorry for the delay, i will update very soon.

Aren't my changes fine?

freddy77:
Merged

PR #397, opened 2021-04-15
Recent changes from NCBI, mostly around portability and thread safety

ucko:
Thanks again for incorporating so much of my previous PR!  I've left that branch as is and established this fresh one for my newer tuneups, which primarily focus on matters of portability and thread safety.

### DIFF
--- a/src/tds/log.c
+++ b/src/tds/log.c
@@ -56,6 +56,10 @@ int tds_g_append_mode = 0;
 static char *g_dump_filename = NULL;
 /** Tell if TDS debug logging is turned on or off */
 int tds_write_dump = 0;
+#ifdef TLS
+TLS
+#endif
+int tdsdump_elided = 0;
 static FILE *g_dumpfile = NULL;	/* file pointer for dump log          */
 static tds_mutex g_dump_mutex = TDS_MUTEX_INITIALIZER;
 
@@ -263,7 +267,8 @@ tdsdump_dump_buf(const char* file, unsigned int level_line, const char *msg, con
 	char line_buf[BYTES_PER_LINE * 8 + 16], *p;
 	FILE *dumpfile;
 
-	if (((tds_debug_flags >> debug_lvl) & 1) == 0 || !tds_write_dump)
+	if (((tds_debug_flags >> debug_lvl) & 1) == 0 || !tds_write_dump
+            || tdsdump_elided)
 		return;
 
 	if (!g_dumpfile && !g_dump_filename)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,7 @@ foreach(fn
 	sys/time.h
 	sys/types.h
 	sys/wait.h
+	threads.h
 	unistd.h
 	fcntl.h
 	wchar.h
@@ -309,6 +310,58 @@ set(CMAKE_REQUIRED_LIBRARIES)
 
 check_struct_has_member("struct tm" "tm_zone" "time.h" HAVE_STRUCT_TM_TM_ZONE)
 config_write("#cmakedefine HAVE_STRUCT_TM_TM_ZONE 1\n\n")
+
+set(TLS NOTFOUND)
+macro(CHECK_TLS_SYNTAX SYNTAX)
+	string(REGEX REPLACE "\\((.*)\\)" "_\\1" SANITIZED ${SYNTAX})
+	string(TOUPPER "HAVE_${SANITIZED}" VARIABLE)
+	string(TOUPPER "CHECK_${SANITIZED}" CHECK)
+	if (NOT DEFINED ${VARIABLE})
+		set(MACRO_CHECK_TLS_DEFINITIONS
+			"-D${CHECK}=1 ${CMAKE_REQUIRED_FLAGS}")
+		message(STATUS "Checking support for TLS syntax ${SYNTAX}")
+		try_compile(${VARIABLE}
+			${CMAKE_BINARY_DIR}
+			${CMAKE_SOURCE_DIR}/misc/cmake_checks.c
+			COMPILE_DEFINITIONS ${CMAKE_REQUIRED_DEFINITIONS}
+			CMAKE_FLAGS
+			-DCOMPILE_DEFINITIONS:STRING=${MACRO_CHECK_TLS_DEFINITIONS}
+			OUTPUT_VARIABLE OUTPUT)
+		if(${VARIABLE})
+			set(${VARIABLE} 1 CACHE INTERNAL
+				"Have TLS syntax ${SYNTAX}")
+			message(STATUS "Checking support for ${SYNTAX}"
+				" - found")
+			file(APPEND
+				${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeOutput.log
+				"Determining if ${SYNTAX} is valid TLS syntax"
+				" passed with the following output:\n"
+				"${OUTPUT}\n\n")
+		else()
+			message(STATUS "Checking support for ${SYNTAX}"
+				" - not found")
+			set(${VARIABLE} "" CACHE INTERNAL
+				"Have TLS syntax ${SYNTAX}")
+			file(APPEND
+				${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log
+				"Determining if ${SYNTAX} is valid TLS syntax"
+				" failed with the following output:\n"
+				"${OUTPUT}\n\n")
+		endif()
+	endif()
+	if(${VARIABLE})
+		set(TLS ${SYNTAX})
+	endif()
+endmacro()
+
+foreach(syntax thread_local _Thread_local __thread "__declspec(thread)")
+	if (${TLS} STREQUAL NOTFOUND)
+		check_tls_syntax(${syntax})
+	endif()
+endforeach(syntax)
+if (NOT ${TLS} STREQUAL NOTFOUND)
+	config_write("#cmakedefine TLS ${TLS}\n\n")
+endif()
 
 macro(SEARCH_LIBRARY FUNC HAVE VAR LIBS)
 	foreach(lib ${LIBS})
--- a/configure.ac
+++ b/configure.ac
@@ -857,6 +857,7 @@ AC_ARG_ENABLE(threadsafe,
 if test "$enable_threadsafe" != "no" ; then
 	CPPFLAGS="$CPPFLAGS -D_REENTRANT -D_THREAD_SAFE"
 	AC_MSG_RESULT(enabled)
+	AX_TLS
 else
 	AC_DEFINE(TDS_NO_THREADSAFE, 1, [Define if you don't care about thread safety])
 	AC_MSG_RESULT(disabled)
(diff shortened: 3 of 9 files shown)


include/freetds/macros.h
```diff
@@ -70,7 +70,7 @@
 #define TDS_WUR
 #endif
 
-#define TDS_INT2PTR(i) ((void*)(((char*)0)+(i)))
+#define TDS_INT2PTR(i) ((void*)(((char*)0)+((intptr_t)(i))))
```
fziglio, line 73: nice, but in some systems intptr_t could be defined by tds_sysdep_public.h which is not included by this file

src/ctlib/ct.c
```diff
@@ -496,10 +496,19 @@ ct_con_props(CS_CONNECTION * con, CS_INT action, CS_INT property, CS_VOID * buff
 		case CS_TIMEOUT:
 			/* set the query timeout as an integer in seconds */
 		        tds_login->query_timeout = *(CS_INT *) buffer;
+			if (tds_login->query_timeout == CS_NO_LIMIT) {
```
fziglio, line 499: style: no bracket for a single line

src/ctlib/unittests/row_count.c
```diff
@@ -122,11 +122,16 @@ test(int final_rows, int no_rows)
 	CS_RETCODE ret;
 	CS_CHAR cmdbuf[4096];
 	char results[1024];
+        char *results_end = results, *cmdbuf_end = cmdbuf;
 
 	run_command(cmd, "create table #tmp1 (i int not null)");
 
-	results[0] = 0;
-	strcpy(cmdbuf, "create proc sample_rpc as ");
+#define strcat(s, k) do { \
```
fziglio, line 129: pretty ugly (maybe more ugly the bug). Why not using strlcpy/strlcat (include replacements.h for portability).

src/odbc/odbc.c
```diff
@@ -1891,7 +1891,9 @@ SQLCancel(SQLHSTMT hstmt)
 
 	tdsdump_log(TDS_DBG_FUNC, "SQLCancel(%p)\n", hstmt);
 
+        tds_mutex_lock(&stmt->dbc->mtx);
```
fziglio, line 1894: not sure if SQLCancel is supposed to be also be called in a signal. This would be a deadlock if already locked.

src/tds/log.c
```diff
@@ -56,6 +56,10 @@ int tds_g_append_mode = 0;
 static char *g_dump_filename = NULL;
 /** Tell if TDS debug logging is turned on or off */
 int tds_write_dump = 0;
+#ifdef TLS
+TLS
+#endif
+int tdsdump_elided = 0;
```
fziglio, line 62: If TLS is not defined this would be a global like `tds_write_dump`

include/freetds/macros.h
ucko: Good point.  Any objection to adding an appropriate `#include` directive?  I see that `tds.h` already includes `tds_sysdep_public.h` ahead of `macros.h`.

src/ctlib/ct.c
ucko: Ah, right, we use different style conventions, and I evidently forgot to flip back to your style here.

include/freetds/macros.h
fziglio: yes, but some files include `macros.h` without including `tds.h`. I suppose it's safe to include `tds_sysdep_public.h` from `macros.h`

src/ctlib/unittests/row_count.c
ucko: This approach takes advantage of known lengths to be marginally faster, but I suppose that's a matter of premature optimization, given that this program is (a) a test and (b) almost certainly I/O bound.  I presume you'd at that point prefer to do away with this `#define` altogether and just call `strlcpy`/`strlcat` directly?

src/odbc/odbc.c
ucko: Good point. :-/  How would you feel about using `tds_mutex_trylock` and (condtionally on `TDS_HAVE_MUTEX`) directly inspecting the `locked_by` field when unable to (re)take the lock?

src/tds/log.c
ucko: Yes, modulo the usual visibility constraints from `pushvis.h`.  That's a necessary evil, given the need to coordinate between `log.c` and `login.c`.

src/tds/unittests/convert.c
```diff
@@ -38,6 +38,7 @@ free_convert(int type, CONV_RESULT *cr)
 {
 	switch (type) {
 	case SYBCHAR: case SYBVARCHAR: case SYBTEXT: case XSYBCHAR: case XSYBVARCHAR:
+	case SYBNVARCHAR: case SYBNTEXT: case XSYBNCHAR: case XSYBNVARCHAR:
```
fziglio, line 41: I honestly don't think tds_convert supports any of these. That's pretty weird. Also valgrind should have catch it. Do you have a reproduction?

src/tds/unittests/convert.c
```diff
@@ -321,8 +322,7 @@ main(int argc, char **argv)
 		 */
 
 		result = tds_convert(ctx, srctype, src, srclen, desttype, &cr);
-		if (result >= 0)
-			free_convert(desttype, &cr);
+		free_convert(desttype, &cr);
```
fziglio, line 325: That change I don't understand. Data should not be allocated if the result is a failure.

src/tds/unittests/convert.c
ucko: In retrospect, this is a fixup for 07a8138 from my other PR, even if it formally applies on its own.

src/tds/unittests/convert.c
ucko: Ah, yeah, this is another, evidently misplaced, fixup for the unmerged 07a8138 from #219.  The odd thing is that the `tds`-layer portions of that change don't add any fully new code, just unblock some options.

src/odbc/odbc.c
ucko: Never mind, `locked_by` is available only under `ENABLE_EXTRA_CHECKS`.

ucko:
I've pushed a new take on this branch with the following changes:
- Rebased against `master` (not that it changed much)
- Corrected `TLS`-related logic in `CMakeLists.txt` that misfired on reconfiguration.
- Revised the patches to `ct.c` and `row_count.c` as discussed.
- Dropped the patches to the `convert.c` test and to `SQLCancel` altogether.

freddy77:
Merged:
- row_count: Work around strcat lossage under LLVM Clang 6 (cleaned up).
- dblib/unittests/common.c: Handle basename and dirname more carefully.
- ct_con_props 1.0: Improve CS_(LOGIN_)TIMEOUT handling.

freddy77:
@ucko, how do you test the code? Just looking at the code I realized it doesn't work, AppVeyor is failing, my tests too (but these are quite extensive so not a big surprise).

freddy77:
By the way, I was thinking a release, I think I won't merge the TLS code in the new release (but plenty of space on master).

freddy77:
@ucko, what do you think about https://github.com/FreeTDS/freetds/commit/39d223d1b233b4672d59ee721537fec8f3017df6 ?

ucko:
Thanks for the (partial) merge!

I primarily tested these changes in NCBI’s 1.0-based branch (overdue for an update, I know), which has two or three tests disabled because they hardcode bad credentials, about which NCBI’s DBAs complained.  (IIRC, they particularly objected to a test that tried to log in as `sa`.)  At any rate, this branch exists in service of `ctlib`-based C++ wrappers with additional tests, including one that’s now heavily multithreaded following some changes to locking and bookkeeping logistics.

Forward-porting these changes seemed straightforward enough that I didn’t retest them as thoroughly as I should have, so some errors and/or outdated assumptions evidently slipped in. :-/ I addressed one such error with my latest fixup to `src/tds/iconv.c`, but see that Travis-CI is still failing; I’ll take another look at what might be going wrong when I get a chance.
As for 39d223d, I mostly agree, but found early on that ThreadSanitizer wanted formal locking around `tds_iconv_open`’s initial `iconv_initialized` check.  Also, `iconv_initialized` was present all along, so modernizing its type to `bool` is orthogonal to any other changes (but does remind me  to  fix a890c3d to use `bool` for `tdsdump_elided`, which is new).

freddy77:
> As for [39d223d](https://github.com/FreeTDS/freetds/commit/39d223d1b233b4672d59ee721537fec8f3017df6), I mostly agree, but found early on that ThreadSanitizer wanted formal locking around `tds_iconv_open`’s initial `iconv_initialized` check. 

Yes, ThreadSanitizer associates variables to mutexes and complaints if the variable is not protected by same mutex. However this is pretty normal pattern for initialization (test without and then with mutex). ThreadSanitizer is not able to understand that the second check fix the issue (coherence) of the first.
Actually I'm not really sure if this patch is really necessary. The only issue not having the mutexes is that the initialization could happen multiple times but I don't think there is any side effect of this, the result will be the same.
In the worst case you will initialize N time with N the number of cores on your system.

> Also, `iconv_initialized` was present all along, so modernizing its type to `bool` is orthogonal to any other changes (but does remind me to fix [a890c3d](https://github.com/FreeTDS/freetds/commit/a890c3dd08d33e7c1734c3665c6a1d3a515c7d74) to use `bool` for `tdsdump_elided`, which is new).

Yes, I started using bool for boolean changing the hunks that I touch.

About the elided patch I was thinking a different implementation not involving TLS but checking thread_id. I was also trying to have different logging setting for each connection but it looks overwhelming.

ucko:
> Yes, ThreadSanitizer associates variables to mutexes and complaints if the variable is not protected by same mutex. However this is pretty normal pattern for initialization (test without and then with mutex). ThreadSanitizer is not able to understand that the second check fix the issue (coherence) of the first.

I concede that it's probably being overly paranoid here, but I'd still favor playing along unless the overhead is a problem in practice.  (Isn't FreeTDS predominantly I/O bound?)

> Yes, I started using bool for boolean changing the hunks that I touch.

Got it, thanks.  Should a890c3d then likewise switch `tds_write_dump` to `bool` along the way?

> About the elided patch I was thinking a different implementation not involving TLS but checking thread_id.

Oh, I see.  I presume the idea would be to allow only one thread to be in (the relevant portion of) `tds7_send_login` at a time?  I'd prefer to avoid holding locks during network I/O, though in this case I suppose it wouldn't be too bad to do so -- AFAICT, this I/O consists of a single small write to a previously established socket, and should therefore amount in practice to a quick handoff to a kernel buffer.

freddy77:
@ucko merged another patch. I was looking at the context one and now that I understood it I think this will go to 1.2 pretty quickly.

ucko:
Great, thanks again!  By my reckoning, that just leaves tds/iconv.c locking, which as discussed is probably safe to leave out after all.

ucko:
Ah, right, also the SQLCancel business, but I don't see a good way to stay fully out of trouble in the multithreaded case without breaking the interrupt-handler case, and don't exercise that API outside of upstream unit tests anyway.

freddy77:
Looking at changes all were implemented in some way or another

PR #399, opened 2021-04-26
Datacopy fix

blieusong:
- Replaced all `SET FMTONLY` by `select top 0` (at least this works with minimal changes to the code)
- Implemented a dirty fix to handle `varchar` better in **tds_get_column_declaration**
- better handling of `(n)varchar` and `(n)char` (still buggy on MSSQL)
- fixed `varchar` handling in bcp


src/apps/datacopy.c
```diff
@@ -411,7 +411,7 @@ create_target_table(char *sobjname, char *owner, char *dobjname, DBPROCESS * dbs
 	DBINT num_cols;
 	DBCOL2 colinfo;
 
-	sprintf(ls_command, "SET FMTONLY ON select * from %s SET FMTONLY OFF", sobjname);
+	sprintf(ls_command, "select top 0 * from %s", sobjname);
```
freddy77, line 414: why not using the more classic `select * from <table> where 0=1` ?

src/apps/datacopy.c
```diff
@@ -445,11 +445,8 @@ create_target_table(char *sobjname, char *owner, char *dobjname, DBPROCESS * dbs
 
 		strlcat(ls_command, colinfo.ServerTypeDeclaration, sizeof(ls_command));
 
-		if (colinfo.Null == TRUE) {
-			strlcat(ls_command, " NULL", sizeof(ls_command));
-		} else {
-			strlcat(ls_command, " NOT NULL", sizeof(ls_command));
-		}
+		strlcat(ls_command, (colinfo.Identity == TRUE) ? " IDENTITY" : " ", sizeof(ls_command));
+		strlcat(ls_command, (colinfo.Null == TRUE) ? " NULL" : " NOT NULL", sizeof(ls_command));
```
freddy77, line 449: Not a regression but I would avoid checking `<bool> == TRUE` and just use `<bool>` or `!!<bool>` if you really want the expression to be bool

src/tds/query.c
```diff
@@ -728,12 +728,48 @@ tds_get_column_declaration(TDSSOCKET * tds, TDSCOLUMN * curcol, char *out)
 	size = tds_fix_column_size(tds, curcol);
 
 	switch (tds_get_conversion_type(curcol->on_server.column_type, curcol->on_server.column_size)) {
-	case XSYBCHAR:
+	case XSYBCHAR: //same as SYBLONGCHAR
+		if (!IS_TDS7_PLUS(tds->conn)) { //SYBLONGCHAR
```
freddy77, line 732: I think this would be better to be `IS_TDS50(tds->conn)`.

src/tds/query.c
```diff
@@ -728,12 +728,48 @@ tds_get_column_declaration(TDSSOCKET * tds, TDSCOLUMN * curcol, char *out)
...
+			switch (curcol->column_usertype) { // dirty temporary hack
```
freddy77, line 733: Please don't use C++ style comments. Although all modern C compilers should support it it's not uncommon that someone use very old limited compilers.

src/tds/query.c
```diff
@@ -728,12 +728,48 @@ tds_get_column_declaration(TDSSOCKET * tds, TDSCOLUMN * curcol, char *out)
...
+				fmt = "CHAR(%u)";
+				break;
+			}
+		}
 		break;
 	case SYBVARCHAR:
-	case XSYBVARCHAR:
+		if (!IS_TDS7_PLUS(tds->conn) && curcol->column_usertype == 25) {
+			if (curcol->column_varint_size == 8)
+				fmt = "NVARCHAR(MAX)";
```
freddy77, line 767: I don't think Sybase supports `nvarchar(max)`

src/tds/bulk.c
```diff
@@ -138,7 +138,17 @@ tds_bcp_init(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
 		 * Or, see if the "lower" part is unused (and zeroed out) at this point, and just do one assignment.
 		 */
 		curcol->funcs = resinfo->columns[i]->funcs;
-		curcol->column_type = resinfo->columns[i]->column_type;
+		/* I believe the whole behaviour of figuring the column type out of column_type is flawed. It should
+		 * be based (at least partially!) on column_usertype (I haven't read documentation but that's from 
+		 * trials and error and the content of systypes table). The below is a dirty hack.
+		 * usertype = 2 > VARCHAR. 25 > NVARCHAR
+		 */
+		if ((resinfo->columns[i]->column_type == SYBCHAR) &&
```
freddy77, line 146: I think this part of the series makes AppVeyor failing pretty bad. I don't really like the same structure to have 2 semantics based on the usage.

src/apps/datacopy.c
blieusong: Because it works fine (on Sybase and MSSQL). What you propose is ANSI compliant (was this your point?), but I'm not sure that's something very important here (we're on Free**TDS** after all).

src/apps/datacopy.c
blieusong: Will fix. Thanks.

src/tds/query.c
blieusong: I agree. Thanks.

src/tds/query.c
blieusong: OK.

src/tds/query.c
blieusong: To be honest I haven't tested that part of the code, and just replicated similar portion. It's that a case which can't occur (col varint size = 8), I can remove it. Or I can default to 32767?

src/tds/bulk.c
blieusong: I'm not sure what you mean and what to do of ur review here... if you are questioning the enum, I agree (two enum for the same value is dirty). Do you propose to split in two? One Sybase enum and another MSSQL enum? I would agree, but maybe one thing at a time? I would also add an enum for the user type which isn't properly handled by FreeTDS as far as I can tell (I haven't fully checked on MSSQL, but it seems to be more of a Sybase issue)

fziglio: I think that this hunk should be removed, the function is just copying a structure, and it should not changing it while copying to simplify another part of the code. About the enum, yes, it became a bit a problem. They are not 2 enums, originally there were not overlapping, then code and versions got forked, so there are some common values, then MS only numbers and Sybase only numbers which, accidentally are now overlapping. The `column_type` (and actually to confuse even move there's a `on_server.column_type`) is the "wire" type which describe how the data is encoded in the protocol. There's also (as you discovered) a `column_usertype` which better describe the type at the user level. For instance the `timestamp` type and `binary` type has the same `column_type` but different `column_usertype`. I think in MS `varchar` and `nvarchar` has different `on_server.column_type` so there should be no problem there. About Sybase probably you are right.

src/tds/query.c
fziglio: Sybase does not support `varchar(max)` or, which is similar (there's also `varbinary(max)`), `column_varint_size` == 8.

src/apps/datacopy.c
fziglio: More than ANSI I think very old MSSQL/ASE databases does not support `TOP` (not sure about ASA either). But as far as I know they should support `SET FMTONLY`.

blieusong: Make sens. It's a free and safe change anyways, so I will do it too. And yes, I remember `select top` was introduced in Sybase 15 (I think?).

src/tds/bulk.c
blieusong: What part exactly would you want removed? If I remove that chunk, the bulk copy won't work.

I do agree that the historic led to that complicated situation.

Surely the code should be rewritten, like separate enum, or even separate functions for each flavor of TDS (I can see it is already done for some functions, but the principle could be expanded). I'm willing to give it a shot, but I don't think it should be on that specific PR. 

I would advocate for incremental approach here, and a rewrite later. Plus, that would give me time to get used to the FreeTDS code (Not that I should be the one rewriting it, but since I'm the one complaining, I'm willing to do it if necessary).

freddy77: I think the error should be fixed in another place. If code is not able to recreate the same type as it was it should be fixed in the code that declare the type, if the code is not able to insert that type of column it should be fixed in the code that insert data. As you said that field is already confusing, adding more meaning to it won't surely improving the situation.

The fact that a `column_type` (and not `on_server.column_type`) solves a problem is already an indication that something at some other level is broken.

Which specific data type are you not able to copy? Is it a copy from different databases (like from Sybase to MSSQL) or just Sybase to Sybase?

About a new enumeration. Do you have an idea of what it should represent? Should they values be independent on type representation? Should be more database "friendly" taking into consideration usertype but not, for instance, if the type is short or long (so having 2 different representation on the row)? Another solution would be possibly to remove it and use just the `on_server` version. On a similar issue there's a function called `tds_get_conversion_type` which `removes` nullable types. Another thing to consider is that dblib is returning a similar enumeration (but reduced and somehow simplified like XML will became a TEXT or not distinguishing between VARCHAR and NVARCHAR). I think this enumeration would be better to be addressed outside the scope of this PR. I would like to prepare a new release shortly, surely this enumeration change won't be there.

fziglio:
I was forgetting. There's a `--enable-extra-checks` options that helps testing internal structures consistencies. Surely not to enable in production but it does a great job, many mistakes are catch when this option is turned on.

src/apps/datacopy.c
blieusong: @freddy77 any reason why Null is a BYTE and not a BOOL?

fziglio: The definition of the original structure came from Microsoft. Possibly they did this decision to save some bytes.

blieusong: In that case I'd keep the `<bool> == TRUE` for the sake of consistency. What do you think ? Would feel weird to see

```c
strlcat(ls_command, (colinfo.Identity) ? " IDENTITY" : " ", sizeof(ls_command));
strlcat(ls_command, (colinfo.Null == TRUE) ? " NULL" : " NOT NULL", sizeof(ls_command));
```

Also to be honest, I've never liked if(value) because `if` treats *any* no zero value to true. `== TRUE` is safer in that regard.

src/tds/bulk.c
blieusong: The reason I fix here *at the moment* is that I'm not sure FreeTDS handle **usertype** correctly. I really don't want to embark on a too long journey and leave that bcp function broken atm. That's up to you. I can already check for a longer term fix (it was my plan anyways), but I would be sad to leave something which doesn't work for Sybase. Thinking about it, many graphical clients which rely on jTDS (and thus, FreeTDS) all display wrong information because of that. dbVisualizer, dBeaver, and probably all of them...

From my tests, it is impossible to guess the column type based on the `on_server` type. It's a surjective transformation where both char and varchar give a XSYBCHAR on the server. So if you start from XSYBCHAR, you are stuck.

I think we should have an enumeration for usertype. One for Sybase and one for MSSQL.

I cannot copy data from Sybase to Sybase (I now forgot my test case...). Maybe I'll continue this discussion on the issue I opened because this is getting quite cluttered now. I will describe my tests in detail there.

blieusong: Also, sorry for my slow replies. I'm unexpectedly busy at work and at home. Hope I can help more soon enough.

src/apps/datacopy.c
blieusong: Fixed.

src/tds/query.c
blieusong: Fixed

src/tds/query.c
blieusong: Fixed

src/apps/datacopy.c
blieusong: I propose not to change that one for the reasons mentioned above.

src/tds/query.c
blieusong: Indeed. Sybase (var)char max length is 16384 (or apparently 16296 depending on pagesize and table lock scheme). I'm not quite sure how to handle this. 

My little book says that this value can be retrieved with **dbcc serverlimits** but that would give a database level limit... also, not everyone can execute **dbcc**.

Is that information carried over by TDS by any chance to your knowledge? I would be surprised if it is.

Should I just stick to 16296?

blieusong:
Hi,

I made some other fixes (not even yet commited) but i'm still strugling with `n(var)char` handling. Freetds doesn't seem to get the datatype length right, at least on Sybase. 

For example if the server charset is utf8, a `NCHAR(300)` is understood as a `NCHAR(900)` because the size of the column is recorded as 900.

I have found that we could get back to the 300 if we take into account the value of `@@ncharsize`.

How would you implement it? I propose that we get the value of `@@ncharsize` just after the connection is established and store it in `struct tds_dblib_dbprocess`.

My guess is that this applies as well to `uni(var)char` types with  `@@unicharsize`.

fziglio:
Let's try doing this step by step. The patch is doing multiple changes. The devcontainer is not strictly related to `datacopy`. The select change can go in a separate commit that we already agree on. The unittests has to pass so we (you and/or me) need to find why they are failing. The main part I think is detecting the exact database type. One issue is that column_type is not enough and we need to consider the usertype (the column_type mainly tells how the data is encoded in the network, not the specific type used in the database). The other, on your last comment, is the length of the type.

The length stored in DBLibrary, as you realized, is not the database length. The reason is that DBLibrary had no concept of Unicode and all lengths are byte lengths. So FreeTDS is converting characters to client charset adjusting length and encoding (nowadays usually converting everything to UTF-8). This is good but for UTF-16 is get the by-3 multiplication you are seeing (note that factor could potentially change depending on client and server/data settings). `@@ncharsize` and `@@unicharsize` are server detail, here we are dealing with client/library details. I don't know if there's a way in DBLibrary (FreeTDS one!) to disable these adjustments. ODBC, to compare, has the support for Unicode and you can get different length for a single column (length in byte, length in characters and database length).

blieusong:
Hi,

The dev container part is in another PR.

As for doing this step by step yes I agree on that. 

I will commit and push the code which handles the longvarchar type from sybase ASE and TDS 5.

As for the ˋ@@ncharsize` issue, I'll check how ˋdbcollen` behave on the official sybase DBlib. If it returns the length in bytes then we needn't fix freetds' implementation and I will submit a fix on datacopy itself. If it returns the number of char on the other hand, we'll have to do something.

fziglio:
Superseded, see `master` branch.

PR #411, opened 2021-08-15
modify for freetds coredump when connect failed

xingqiaojin:
modify for freetds coredump when connect failed


fziglio:
It's not clean what this PR is about.

xingqiaojin:
> It's not clear what this PR is about.
If it returned INT_EXIT when connect the DB is failed , will cause the service restart. so we need modify it.

freddy77:
You have to set the error handler

PR #412, opened 2021-08-16
Keep make from failing if the source path contains spaces.

marsupilami79:
Currently when building freetds the build fails if the build path has spaces in its name. This fixes the issue for me.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -112,7 +112,7 @@ $(DOCDIR)/userguide/index.html: 	$(DOCDIR) userguide.xml userguide_desc.xml \
 		fi; \
 	fi
 # try xmlto ...
-	test -r $(abs_srcdir)/userguide.xml
+	test -r "$(abs_srcdir)/userguide.xml"
 	test -r userguide.xml || ln -s $(srcdir)/userguide.xml userguide.xml
 	test -r userguide.xsl || ln -s $(srcdir)/userguide.xsl userguide.xsl
 	test -r userguide_desc.xml || ln -s $(srcdir)/userguide_desc.xml userguide_desc.xml
@@ -120,7 +120,7 @@ $(DOCDIR)/userguide/index.html: 	$(DOCDIR) userguide.xml userguide_desc.xml \
 	if xmlto --version > /dev/null 2> /dev/null ; then \
 		xmlto -m ../userguide.xsl xhtml ../userguide.xml; \
 ## 		and copy the stylesheet
-		cp $(abs_srcdir)/userguide.css .; \
+		cp "$(abs_srcdir)/userguide.css" .; \
 	fi)
 # if still no joy, create a stub
 	if test ! -r $@ ; then \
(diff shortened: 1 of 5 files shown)


doc/Makefile.am
```diff
@@ -112,7 +112,7 @@ $(DOCDIR)/userguide/index.html: 	$(DOCDIR) userguide.xml userguide_desc.xml \
 		fi; \
 	fi
 # try xmlto ...
-	test -r $(abs_srcdir)/userguide.xml
+	test -r "$(abs_srcdir)/userguide.xml"
```
fziglio, line 115: I suppose the lines below should be updated too

marsupilami79: Honestly I am not sure because I only guessed at what to change. ;)

For me the build works without further changes. I can only assume this has to do with one line using abs_srcdir and the other line using srcdir?

Addendum: I added some echos to the makefile. During the build srcdir is
`.`
while abs_srcdir is
`/home/jan/freetds 1.3/freetds/doc`.
This is why the other lines don't need extra quotes:

~~~
echo "-------------"
-------------
echo "abs_srcdir: /home/jan/freetds 1.3/freetds/doc"
abs_srcdir: /home/jan/freetds 1.3/freetds/doc
echo "srcdir: ."
srcdir: .
echo "-------------"
-------------
test -r "/home/jan/freetds 1.3/freetds/doc/userguide.xml"
test -r userguide.xml || ln -s ./userguide.xml userguide.xml
test -r userguide.xsl || ln -s ./userguide.xsl userguide.xsl
test -r userguide_desc.xml || ln -s ./userguide_desc.xml userguide_desc.xml
~~~

I am not sure, why abs_srcdir gets used instead of srcdir.

fziglio:
I see. I think if the parent has a space this should fix, if between the build and the configure you have a space probably this patch is not enough.

marsupilami79:
I am not sure, what you mean. Since the FreeTDS sources don't contain any paths with spaces, this should be no issue?
If you want me to add more quotes, this is no problem. Just let me know :)

fziglio:
No, I was not asking for it. But while we are at it maybe just quote the `abs_`*

```
$ find -name Makefile.am | xargs grep abs
./doc/Makefile.am:	test -r $(abs_srcdir)/userguide.xml
./doc/Makefile.am:		cp $(abs_srcdir)/userguide.css .; \
./src/dblib/unittests/Makefile.am:AM_LDFLAGS	=	-no-install -L../.libs -R $(abs_builddir)/../.libs
./src/tds/unittests/Makefile.am:AM_LDFLAGS	=	-no-install -L../.libs -R $(abs_builddir)/../.libs
./src/ctlib/unittests/Makefile.am:AM_LDFLAGS	=	-no-install -L../.libs -R $(abs_builddir)/../.libs
./src/odbc/unittests/Makefile.am:AM_LDFLAGS	=	-no-install -L../.libs -R $(abs_builddir)/../.libs
```

marsupilami79:
I changed the files. Unfortunately I made some mistake and couldn't get git to amend the changes to the first commit :(

PR #420, opened 2021-09-13
odbc: connection-string allow any freetds.conf option

Kokan:
The aim is to be transparent, and allow any freetds.conf configuration option in the ODBC connection-string, so user do not require freetds.conf string.


fziglio:
The patch seems sensible. Some considerations:
- the options in ODBC are case insensitive, the option in libTDS are case sensitive, I think he easiest fix for this inconsistency would be to have all case insentitive;
- there's no check for function result, I think we should consider `valid_configuration` field in `odbc_parse_connect_string`;
- `odbc_parse_connect_string` fills `parsed_params` for the supported ODBC parameters but not for libTDS options. This feature is used to fill `ConnStrOut` in `SQLDriverConnect` allowing to expand the connection string to a full independent connection string. I think we can ignore this but it would be better if we could support this too.

Kokan:
Thanks for the feedback. I'll update the PR with your points.

mmcnabb-vms:
@Kokan just looking at this old PR... did you ever make any updates?

PR #449, opened 2022-01-19
More NCBI changes for your consideration (Jan. 2022)

ucko:
Happy new year!  I have four more patches for your consideration; I've rolled them up into a single PR, but they're mutually independent, so you should be able to take or leave them individually.  (Likewise for 27c7a831's changes to each individual file.)

### DIFF
--- a/src/tds/convert.c
+++ b/src/tds/convert.c
@@ -78,26 +78,43 @@ static const char *parse_numeric(const char *buf, const char *pend,
 
 #define test_alloc(x) {if ((x)==NULL) return TDS_CONVERT_NOMEM;}
 
-#define IS_TINYINT(x) ( 0 <= (x) && (x) <= 0xff )
-#define IS_SMALLINT(x) ( -32768 <= (x) && (x) <= 32767 )
-#define IS_USMALLINT(x) ( 0 <= (x) && (x) <= 65535 )
 /*
+ * Macros for number checks.
+ * They work for signed/unsigned integers and floating point values
+ * (accounting for truncation of any fractional portions of the latter).
+ */
+#define IS_TINYINT(x) ( -1 < (x) && (x) < 0x100 )
+#define IS_SMALLINT(x) ( -32769 < (x) && (x) < 32768 )
+#define IS_USMALLINT(x) ( -1 < (x) && (x) < 65536 )
+
+/*
+ * Macros for integer number checks.
+ * IS_UINT works for both integers and floating point values.
+ *
  * f77: I don't write -2147483648, some compiler seem to have some problem 
  * with this constant although is a valid 32bit value
  */
-#define TDS_INT_MIN (-2147483647l-1l)
-#define TDS_INT_MAX 2147483647l
-#define IS_INT(x) (TDS_INT_MIN <= (x) && (x) <= TDS_INT_MAX)
-
-#define TDS_UINT_MAX 4294967295lu
-#define IS_UINT(x) (0 <= (x) && (x) <= TDS_UINT_MAX)
-
-#define TDS_INT8_MAX ((((TDS_INT8) 0x7fffffffl) << 32) + (TDS_INT8) 0xfffffffflu)
+#define TDS_INT_MIN (-2147483647-1)
+#define TDS_INT_MAX 2147483647
+#define INT_IS_INT(x) (TDS_INT_MIN <= (x) && (x) <= TDS_INT_MAX)
+#define TDS_UINT_MAX 4294967295u
+#define IS_UINT(x) (-1 < (TDS_INT8)(x) && (x) < (TDS_INT8) TDS_UINT_MAX + 1)
 #define TDS_INT8_MIN (-(((TDS_INT8)1)<<62) -(((TDS_INT8)1)<<62))
-#define IS_INT8(x) (TDS_INT8_MIN <= (x) && (x) <= TDS_INT8_MAX)
+#define TDS_INT8_MAX ((((TDS_INT8) 0x7fffffff) << 32) + (TDS_INT8) 0xffffffffu)
 
-#define TDS_UINT8_MAX ((((TDS_UINT8) 0xfffffffflu) << 32) + 0xfffffffflu)
-#define IS_UINT8(x) (0 <= (x) && (x) <= TDS_UINT8_MAX)
+/*
+ * Macros for floating point number checks.
+ * To avoid possible loss of precision, use upper bounds for floating point
+ * values and test MIN - (x) < 1 rather than MIN - 1 < (x) for signed
+ * destination types.
+ */
+#define TDS_INT_UPPER_FLOAT 2147483648.0f
+#define FLOAT_IS_INT(x) (TDS_INT_MIN - (x) < 1.0f && (x) < TDS_INT_UPPER_FLOAT)
+#define TDS_INT8_MIN_FLOAT (-9223372036854775808.0f)
+#define TDS_INT8_UPPER_FLOAT 9223372036854775808.0f
+#define FLOAT_IS_INT8(x) (TDS_INT8_MIN_FLOAT - (x) < 1.0f && (x) < TDS_INT8_UPPER_FLOAT)
+#define TDS_UINT8_UPPER 18446744073709551616.0f
+#define FLOAT_IS_UINT8(x) (-1.0f < (x) && (x) < TDS_UINT8_UPPER)
 
 #define TDS_ISDIGIT(c) ((c) >= '0' && (c) <= '9')
 
@@ -446,7 +463,7 @@ tds_convert_char(const TDS_CHAR * src, TDS_UINT srclen, int desttype, CONV_RESUL
 			cr->m.mny = mymoney;
 			return sizeof(TDS_MONEY);
 		} else {
-			if (!IS_INT(mymoney))
+			if (!INT_IS_INT(mymoney))
 				return TDS_CONVERT_OVERFLOW;
 			cr->m4.mny4 = (TDS_INT) mymoney;
 			return sizeof(TDS_MONEY4);
@@ -713,7 +730,7 @@ tds_convert_int8(const TDS_INT8 *src, int desttype, CONV_RESULT * cr)
 	TDS_CHAR tmp_str[24];
 
 	memcpy(&buf, src, sizeof(buf));
-	if (IS_INT(buf))
+	if (INT_IS_INT(buf))
 		return tds_convert_int((TDS_INT) buf, desttype, cr);
 
 	switch (desttype) {
@@ -790,8 +807,8 @@ tds_convert_uint8(const TDS_UINT8 *src, int desttype, CONV_RESULT * cr)
 	TDS_CHAR tmp_str[24];
 
 	memcpy(&buf, src, sizeof(buf));
-	/* IS_INT does not work here due to unsigned/signed conversions */
-	if (buf <= (TDS_UINT8) TDS_INT_MAX)
+	/* INT_IS_INT does not work here due to unsigned/signed conversions */
+	if (buf <= TDS_INT_MAX)
 		return tds_convert_int((TDS_INT) buf, desttype, cr);
 
 	switch (desttype) {
@@ -815,7 +832,7 @@ tds_convert_uint8(const TDS_UINT8 *src, int desttype, CONV_RESULT * cr)
 		return sizeof(TDS_UINT);
 		break;
 	case SYBINT8:
-		if (buf > (TDS_UINT8) TDS_INT8_MAX)
+		if (buf > TDS_INT8_MAX)
 			return TDS_CONVERT_OVERFLOW;
 		cr->bi = (TDS_INT8) buf;
 		return sizeof(TDS_INT8);
@@ -1195,7 +1212,7 @@ tds_convert_money(const TDSCONTEXT * tds_ctx, const TDS_MONEY * src, int desttyp
 		break;
 	case SYBINT4:
 		dollars = mymoney / 10000;
-		if (!IS_INT(dollars))
+		if (!INT_IS_INT(dollars))
 			return TDS_CONVERT_OVERFLOW;
 		cr->i = (TDS_INT) dollars;
 		return sizeof(TDS_INT);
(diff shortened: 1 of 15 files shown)


src/ctlib/unittests/rpc_ct_setparam.c
```diff
@@ -414,14 +415,14 @@ int i, j;
 				coldata[i].value[0] = 0;
 				if (coldata[i].value == NULL) {
 					fprintf(stderr, "malloc coldata.value failed \n");
-					return 1;
+					break;
```
fziglio, line 418: I think you need also to set ret to `CS_FAIL`, otherwise you won't report an error

src/tds/convert.c
```diff
@@ -1530,13 +1530,13 @@ tds_convert_real(const TDS_REAL* src, int desttype, CONV_RESULT * cr)
 		return sizeof(TDS_USMALLINT);
 		break;
 	case SYBINT4:
-		if (!IS_INT(the_value))
+		if (!IS_INT((double)the_value))
```
fziglio, line 1533: nice catch, not sure however if this is really fixing an issue or just hiding a warning. It fixes loosing precision in the conversion but is it possible that it does not catch a number which could not be converted to int, or, otherwise, that it returns overflow for a number that can be converted to int?

src/ctlib/unittests/rpc_ct_setparam.c
ucko: I'll take another look, thanks.  Same goes for `lang_ct_param.c`, I suppose?

src/tds/convert.c
ucko: As https://godbolt.org/z/Ea18E5zTM illustrates, the issue is that due to limited `float` precision, you can't get too close to the limit without going slightly *over* it, at which point conversion will wrap around rather than saturating.

fziglio: It's more about rounding and precision. I was just wondering what's the best way without having to do conversions. Also I was looking at same function trying to understand if the same issue could affect other conversions.

fziglio:
Merged
- ` _ct_handle_client_message: Report appropriate severity values` (minor space changes)
- `ct_describe: For CS_BIGDATETIME_TYPE, supply the server type in usertype`

src/ctlib/unittests/rpc_ct_setparam.c
fziglio: > I'll take another look, thanks. Same goes for `lang_ct_param.c`, I suppose?

where exactly?

src/tds/convert.c
ucko: Good question.  Upon further consideration, how about leaving these call sites alone and instead tweaking the macros?  AFAICT,
```
#define IS_INT(x) (TDS_INT_MIN <= (x) && (x) < (int64_t) TDS_INT_MAX + 1)
/* ... */
#define IS_UINT(x) (0 <= (x) && (x) < (uint64_t) TDS_UINT_MAX + 1)
```
should impose precisely correct limits.  They will let through `double` values exceeding `TDS_(U)INT_MAX` by less than one, but I'd argue that that's actually appropriate given that conversion is defined to round down anyway.

fziglio: Agreed. I would put all signed in the macros like
```c
#define IS_INT(x) (TDS_INT_MIN <= (x) && (x) < (int64_t) TDS_INT_MAX + 1)
/* ... */
#define IS_UINT(x) (0 <= (x) && (x) < (int64_t) TDS_UINT_MAX + 1)
```
not sure about `int64_t`/`TDS_INT8`, maybe a future style update for the other parts

ucko: I'm on board with the all-signed version.  64-bit types are trickier, particularly on the unsigned front, but AFAICT the only use of `IS_(U)INT8` is on a `long` (line 1116), where it amounts to a simple non-negativity check -- `tds_convert_real` and `tds_convert_flt8` use hand-coded boundary checks that attempt to test directly against `_MAX` values.  I'll prepare a revised patch.

src/ctlib/unittests/rpc_ct_setparam.c
ucko: Never mind, I misremembered which pair of tests wound up having parallel fixes.

src/tds/convert.c
```diff
@@ -1538,13 +1538,13 @@ tds_convert_real(const TDS_REAL* src, int desttype, CONV_RESULT * cr)
...
 	case SYBINT8:
-		if (the_value > (TDS_REAL) TDS_INT8_MAX || the_value < (TDS_REAL) TDS_INT8_MIN)
+		if (!IS_INT8(the_value))
 			return TDS_CONVERT_OVERFLOW;
 		cr->bi = (TDS_INT8) the_value;
 		return sizeof(TDS_INT8);
 		break;
 	case SYBUINT8:
-		if (the_value > (TDS_REAL) TDS_UINT8_MAX || the_value < 0)
+		if (the_value >= TDS_UINT8_MAX + 1.0f || the_value < 0)
```
fziglio, line 1547: Wondering if this `TDS_UINT8_MAX + 1.0f` is portable or not. Technically the compiler will convert the first number to float then do the addition. However nor float nor double have the precision to hold all the number precisely (double has 54 bit precision while he number is 63 bit precise). It will probably round directly the number, then add 1 (which won't change). Potentially the result is what we are expecting, but hard to say. If the compile would do truncation on the first conversion the result is not the expected number.

fziglio: Why not having something like
```c
#define TDS_UINT8_UPPER 18446744073709551616.0
```

src/tds/convert.c
```diff
@@ -78,23 +78,23 @@ static const char *parse_numeric(const char *buf, const char *pend,
...
+#define IS_INT(x) (TDS_INT_MIN <= (x) && (x) < (int64_t) TDS_INT_MAX + 1)
 
 #define TDS_UINT_MAX 4294967295lu
-#define IS_UINT(x) (0 <= (x) && (x) <= TDS_UINT_MAX)
+#define IS_UINT(x) (0 <= (x) && (x) < (int64_t) TDS_UINT_MAX + 1)
 
 #define TDS_INT8_MAX ((((TDS_INT8) 0x7fffffffl) << 32) + (TDS_INT8) 0xfffffffflu)
 #define TDS_INT8_MIN (-(((TDS_INT8)1)<<62) -(((TDS_INT8)1)<<62))
-#define IS_INT8(x) (TDS_INT8_MIN <= (x) && (x) <= TDS_INT8_MAX)
+#define IS_INT8(x) (TDS_INT8_MIN <= (x) && (x) < (uint64_t) TDS_INT8_MAX + 1)
```
fziglio, line 97: It should be fine... but `IS_INT8(-1)` is false. It's fine in the sense that we should use this to check unsigned numbers and floating point where the check is correct. But if this is used for signed numbers it will fail.

fziglio:
Merged static analyzer commit. I tried to fix the possible truncating issue, see https://github.com/freddy77/freetds/commit/b0b8887ce6d254a6e04773205fa6a71cc13a87b7 (on top of your, potentially a fixup commit).

src/tds/convert.c
ucko: From what I gather (details below), the `TDS_UINT8_MAX` approach should be pretty portable.  Upon consideration, though, a third possibility could be to take advantage of the fact that `TDS_UINT8_MIN` should have an exact representation by multiplying it by `-2.0` -- per https://stackoverflow.com/questions/29174165/is-flt-radix-ever-not-2-in-c11-for-modern-general-purpose-computers, in the rare event that `FLT_RADIX` isn't two, it should still be a power of two, which suffices (with decimal floating-point, if supported, having separate `_DecimalN` types).

As for rounding, per https://en.cppreference.com/w/c/language/conversion#Real_floating-integer_conversions, on implementations supporting IEEE floating point (which should be more or less all of them), rounding is to nearest for anything evaluable at compile time unless `FENV_ACCESS` (https://en.cppreference.com/w/c/preprocessor/impl) is `ON` (unusual) *and* the application has used `fesetround` to specify some other rounding mode.

Meanwhile, I'd forgotten that the first step in conversion from floating point was truncation rather than taking a floor, so I'll want to allow more margin on signed lower bounds accordingly.

src/tds/convert.c
ucko: Good point.  I suppose an explicit comparison with zero is in order, though doing it up front should at least keep the total number of comparisons at two:
`#define IS_INT8(x) (((x) < 0) ? TDS_INT8_MIN <= (x) : (x) < (uint64_t) TDS_INT8_MAX + 1)`

src/tds/convert.c
```diff
@@ -78,26 +78,37 @@ static const char *parse_numeric(const char *buf, const char *pend,
...
-#define IS_USMALLINT(x) ( 0 <= (x) && (x) <= 65535 )
+/*
+ * Confirm convertability to a signed type with the given bounds,
+ * bearing in mind that promotion rules can complicate comparison
+ * between negative and unsigned integers, that conversion from
+ * floating-point values to integers starts by truncating fractional
+ * portions, and that lb - 1 might have the same FP representation as
+ * lb in x's type (as with float -> int and anything -> TDS_INT8).
+ */
+#define IS_INTn(x, lb, ub) ((x) < 0 ? lb - (x) < 1 : (x) < ub + 1)
```
fziglio, line 89: To be honest this looks complicated... and also I don't think compilers can reach the level of the 2 initial simple comparisons. And the `lb - (x)` part could be subject to loose of precision.

src/tds/convert.c
```diff
@@ -78,26 +78,37 @@ static const char *parse_numeric(const char *buf, const char *pend,
...
-#define TDS_INT8_MAX ((((TDS_INT8) 0x7fffffffl) << 32) + (TDS_INT8) 0xfffffffflu)
+#define TDS_INT8_MAX ((((TDS_INT8) 0x7fffffff) << 32) + (TDS_INT8) 0xffffffffu)
 #define TDS_INT8_MIN (-(((TDS_INT8)1)<<62) -(((TDS_INT8)1)<<62))
-#define IS_INT8(x) (TDS_INT8_MIN <= (x) && (x) <= TDS_INT8_MAX)
+#define IS_INT8(x) IS_INTn(x, TDS_INT8_MIN, (TDS_UINT8) TDS_INT8_MAX)
 
-#define TDS_UINT8_MAX ((((TDS_UINT8) 0xfffffffflu) << 32) + 0xfffffffflu)
+#define TDS_UINT8_MAX ((((TDS_UINT8) 0xffffffffu) << 32) + 0xffffffffu)
 #define IS_UINT8(x) (0 <= (x) && (x) <= TDS_UINT8_MAX)
+#define FLOAT_IS_UINT8(x) ((x) < -2.0f * TDS_INT8_MIN)
```
fziglio, line 111: I think you forgot to check for negativity.

fziglio:
Overall I don't know what to suggest. Macros are useful if easy and reused, but if they have so much side effects expanded in a context or in another I don't know if they make much sense at this point. Another thing I would consider is a test to check the various boundaries.

src/tds/convert.c
ucko: Oops, so I did.

ucko:
Yeah, a test is probably a good idea; should it go to the existing `src/tds/unittests/convert.c`, or should I introduce a `convert2.c` or the like?

src/tds/convert.c
ucko: The general idea with this macro is to unify cases that would otherwise require slightly different pairs of comparisons.  In particular:
- For `int`, `(x) >= TDS_INT_MIN` is correct for every source type but `double`, where it misses the open interval `(TDS_INT_MIN - 1, TDS_INT_MIN)`, and `(x) > (TDS_INT8) TDS_INT_MIN - 1` is correct for every source type but `float`, where `TDS_INT_MIN - 1` is unrepresentable and typically rounds to `TDS_INT_MIN`.  Looking at `lb - (x)` should yield correct results in either case, but will misfire if `x` is `unsigned int` or `TDS_UINT8`, necessitating upfront comparison with zero to be fully general.
- For `TDS_INT8`, `(x) <= TDS_INT8_MAX` picks up floating-point values that are slightly out of range and will wrap around upon conversion; `(x) < (TDS_UINT8) TDS_INT8_MAX + 1` fixes that but proceeds to reject negative integer values, likewise necessitating upfront comparison with zero to be fully general.

At any rate, I'm comparing `lb - (x) < 1` rather than `lb - 1 < (x)` due to precision concerns -- if `lb - 1` loses precision, the comparison could (and typically would) wrongly exclude `lb`, whereas `lb - (x)` will lose precision only if the difference is very positive (and hence clearly greater than one) or very negative (and hence clearly less than one).

fziglio:
I finally manage to write the test. See https://github.com/FreeTDS/freetds/commit/ba2c5e2a152c6b980cd1beebba6f5947acc89510 and https://github.com/FreeTDS/freetds/commit/1ad2a894d4b0c1e8d1a6e3b7a282e73005101d9c.

ucko:
Nice, thanks!  This test passes with my previously submitted changes plus that missing comparison with zero; I've rebased my branch accordingly and pushed it.

src/tds/convert.c
fziglio: I tried something similar under compiler explorer:
```c++
#define INSIDE(x) ((x) > -4 && (x) < 10)

bool limit1(int a) {
    return INSIDE(a);
}

bool limit2(double a) {
    return INSIDE(a);
}

#define IS_INTn(x, lb, ub) ((x) < 0 ? lb - (x) < 1 : (x) < ub + 1)
#define INSIDE2(a) IS_INTn(a, -4, 10)

bool limit3(int a) {
    return INSIDE2(a);
}

bool limit4(double a) {
    return INSIDE2(a);
}
```
and in all compilers I tried the code is less efficient than normal bound checks (as I expected).
I wrote this https://github.com/freddy77/freetds/commit/7ebc5d85e6c001838bd2f6cba4bd623f5e5eea48 to simplify and use simpler bound checks. Please consider to squash it in.

ucko: Squashed, thanks, with some fine-tuning to account for pre-conversion truncation of negative FP values (which I'd previously missed myself for unsigned destination types).

fziglio:
Merged. Wondering however if the test is completely testing correctly bounds for floating points

ucko:
Molto grazie!  On the test front, perhaps you could make judicious use of `nextafter` and `nextafterf`.  On the upper end, the first out-of-range value will be some (exactly representable) power of two `x`, with `nextafter[f](x, 0)` barely making the cut.  The lower end for unsigned types is likewise straightforward -- `-1` is out, but `nextafter[f](-1, 0)` should be legitimate.  For signed types, it's alas not quite so simple -- if `x` is a floating point value corresponding to the lowest in-range integer (the negative of a power of two), then you can check whether (e.g.) `x-1 == x` to determine whether the cutoff is between `nextafter[f](x, 2*x)` and `x` or between `x-1` and `nextafter[f](x-1, 0)`.

PR #457, opened 2022-02-20
Devcontainer

blieusong:
Hi, I would like to resume my work on PR #399 so I have added devcontainer configuration files to the project.

It should be harmless, and would allow be to spin off a FreeTDS dev environnement very quickly, at home, on Github Codespaces, or on Gitpod.

Also made a tiny fix to a .gitignore file.


fziglio:
The series looks still work in progress. The commit https://github.com/FreeTDS/freetds/pull/457/commits/c86cd08d867b75d25a246aff927cf526e0879698 looks good beside the message, at least something like `Ignore autogenerated sysdep_types.h`.

The devcontainer setup looks a bit personal. For instance most of containers are on your account.

blieusong:
Hi, It's quite functional. I'm using it without hiccups so far.

I use my Docker account because I don't have alternatives. If it's fine with you it's possible to create a freetds account. Then the docker images would be built with the latest freetds releases?

I'm also not aware of official Sybase ASE images. Hence why I'm using one I made. It's functional as well, but you can review it if you are curious about how it's made. https://github.com/blieusong/sybase-ase-docker. This could also be moved under a more official umbrella if you think it works well enough.

In short, this is very convenient for testing and debugging purposes (I'm finding many issues with n(var)char and unichar datatypes).

fziglio:
Not saying it's not functional. However is not bound to project, nor with account not with support. What about a README file in the directory (maybe in markdown notation) stating that this is not official? I know docker containers are useful (I'm currently working on the "cloud") but I found security is a bit poor. Not saying that you could inject malicious code in the containers but some people could do if they hack your account, or if they hack docker hub account (I think report about hacking accounts are reported monthly these days).

PR #460, opened 2022-04-20
Allow embedded braces in connection-string values

OverOrion:
We've encountered an issue with escaping: if the value field contains the `;` character then it is needed to be escaped with braces as mentioned in the documentation: https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqldriverconnect-function?view=sql-server-ver15#comments

However there weren't any mention regarding embedded braces in value fields so if `key={val};ue}` then it was parsed as `key={val}`. 
The escape mechanism is not mentioned in the docs, but it seems like that doubling the ending braces (i. e. `key={value}};ue}`) is the way to go: https://github.com/mkleehammer/pyodbc/issues/569#issuecomment-496234942

This PR implements this behaviour and adds tests regarding these scenarios.

### DIFF
--- a/src/odbc/connectparams.c
+++ b/src/odbc/connectparams.c
@@ -268,6 +268,65 @@ odbc_dstr_swap(DSTR *a, DSTR *b)
 	*b = tmp;
 }
 
+static const char *
+parse_value(TDS_ERRS *errs, const char *p, const char *connect_string_end, DSTR *value)
+{
+	const char *end;
+	char *dst;
+
+	/* easy case, just ';' terminated */
+	if (p == connect_string_end || *p != '{') {
+		end = (const char *) memchr(p, ';', connect_string_end - p);
+		if (!end)
+			end = connect_string_end;
+		if (!tds_dstr_copyn(value, p, end - p)) {
+			odbc_errs_add(errs, "HY001", NULL);
+			return NULL;
+		}
+		return end;
+	}
+
+	++p;
+	/* search "};" */
+	end = p;
+	for (;;) {
+		/* search next '}' */
+		end = (const char *) memchr(end, '}', connect_string_end - end);
+		if (end == NULL) {
+			odbc_errs_add(errs, "HY000", "Syntax error in connection string");
+			return NULL;
+		}
+		end++;
+
+		/* termination ? */
+		if (end == connect_string_end || end[0] == ';') {
+			end--;
+			break;
+		}
+
+		/* wrong syntax ? */
+		if (end[0] != '}') {
+			odbc_errs_add(errs, "HY000", "Syntax error in connection string");
+			return NULL;
+		}
+		end++;
+	}
+	if (!tds_dstr_alloc(value, end - p)) {
+		odbc_errs_add(errs, "HY001", NULL);
+		return NULL;
+	}
+	dst = tds_dstr_buf(value);
+	for (; p < end; ++p) {
+		char ch = *p;
+		*dst++ = ch;
+		if (ch == '}')
+			++p;
+	}
+	tds_dstr_setlen(value, dst - tds_dstr_buf(value));
+
+	return end + 1;
+}
+
 /** 
  * Parse connection string and fill login according
  * @param connect_string     connect string
@@ -316,25 +375,9 @@ odbc_parse_connect_string(TDS_ERRS *errs, const char *connect_string, const char
 
 		/* parse value */
 		p = end + 1;
-		if (*p == '{') {
-			++p;
-			/* search "};" */
-			end = p;
-			while ((end = (const char *) memchr(end, '}', connect_string_end - end)) != NULL) {
-				if ((end + 1) != connect_string_end && end[1] == ';')
-					break;
-				++end;
-			}
-		} else {
-			end = (const char *) memchr(p, ';', connect_string_end - p);
-		}
+		end = parse_value(errs, p, connect_string_end, &value);
 		if (!end)
-			end = connect_string_end;
-
-		if (!tds_dstr_copyn(&value, p, end - p)) {
-			odbc_errs_add(errs, "HY001", NULL);
 			return false;
-		}
 
 #define CHK_PARAM(p) (strcasecmp(option, odbc_param_##p) == 0 && (num_param=ODBC_PARAM_##p) >= 0)
 		if (CHK_PARAM(Server)) {
@@ -458,11 +501,9 @@ odbc_parse_connect_string(TDS_ERRS *errs, const char *connect_string, const char
 			odbc_dstr_swap(dest_s, &value);
 
 		p = end;
-		/* handle "" ";.." "};.." cases */
+		/* handle "" ";.." cases */
 		if (p >= connect_string_end)
 			break;
-		if (*p == '}')
-			++p;
 		++p;
 	}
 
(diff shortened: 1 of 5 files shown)


fziglio:
@OverOrion, I saw some changes due to some issues. What is the current status? I mean, it's ready to get reviewed?

OverOrion:
Hi @fziglio,

Yes, I needed to fix Windows specific things, it's ready for review, thank you!

src/odbc/connectparams.c
```diff
@@ -268,6 +286,205 @@ odbc_dstr_swap(DSTR *a, DSTR *b)
 	*b = tmp;
 }
 
+static bool
+odbc_parse_connect_string_value_escaped(TDS_ERRS *errs, const char **current_position, const char **current_end, const char *connect_string_end, DSTR *value)
+{
+	const char *end;
+	char *new_value = strdup(*current_position);
```
fziglio, line 293: `strdup` can fail. Why do you need to call it by the way? Won't it be easier to duplicate all at the beginning and make the string writable?

src/odbc/connectparams.c
```diff
@@ -268,6 +286,205 @@ odbc_dstr_swap(DSTR *a, DSTR *b)
...
+	int num_param = -1;
+	enum { CFG_DSN = 1, CFG_SERVER = 2, CFG_SERVERNAME = 4 };
+
+#define CHK_PARAM(p) (strcasecmp(key, odbc_param_##p) == 0 && (num_param=ODBC_PARAM_##p) >= 0)
+	if (CHK_PARAM(Server)) {
+		/* error if servername or DSN specified */
+		if ((*cfgs & (CFG_DSN|CFG_SERVERNAME)) != 0) {
+			tds_dstr_free(value);
+			odbc_errs_add(errs, "HY000", "Only one between SERVER, SERVERNAME and DSN can be specified");
+			return false;
```
fziglio, line 372: This is not a value in the expected enumeration.

src/odbc/connectparams.c
```diff
@@ -280,19 +497,18 @@ odbc_parse_connect_string(TDS_ERRS *errs, const char *connect_string, const char
 			  TDS_PARSED_PARAM *parsed_params)
 {
 	const char *p, *end;
-	DSTR *dest_s, value = DSTR_INITIALIZER;
-	enum { CFG_DSN = 1, CFG_SERVER = 2, CFG_SERVERNAME = 4 };
+	DSTR value = DSTR_INITIALIZER;
 	unsigned int cfgs = 0;	/* flags for indicate second parse of string */
 	char option[24];
 	int trusted = 0;
+	enum SetParamResult result = 0;
```
fziglio, line 504: 0 is not a value in the enumeration

src/odbc/connectparams.c
```diff
@@ -280,19 +497,18 @@ odbc_parse_connect_string(TDS_ERRS *errs, const char *connect_string, const char
...
+	const char *yes_value = "Yes";
 
-	if (parsed_params)
-		memset(parsed_params, 0, sizeof(*parsed_params)*ODBC_PARAM_SIZE);
+	if (parsed_params) {
+		tds_parsed_param_init(parsed_params);
+	}
```
fziglio, line 509: style: no bracket for a single statement

src/odbc/odbc.c
```diff
@@ -615,31 +615,37 @@ ODBC_FUNC(SQLDriverConnect, (P(SQLHDBC,hdbc), P(SQLHWND,hwnd), PCHARIN(ConnStrIn
...
-			params[ODBC_PARAM_UID].p = NULL;
-			params[ODBC_PARAM_PWD].p = NULL;
-			params[ODBC_PARAM_Trusted_Connection].p = "Yes";
-			params[ODBC_PARAM_Trusted_Connection].len = 3;
+			tds_dstr_empty(&params[ODBC_PARAM_UID]);
+			tds_dstr_empty(&params[ODBC_PARAM_PWD]);
+			if(!tds_dstr_copyn(&params[ODBC_PARAM_Trusted_Connection], yes_value, strlen(yes_value))) {
+        tds_parsed_param_free(params);
+        ODBC_EXIT_(dbc);
+      }
```
fziglio, line 637: style: it's wrongly indented

src/odbc/unittests/connection_string_parse.c
```diff
@@ -0,0 +1,198 @@
...
+
+
+#ifdef _WIN32
+HINSTANCE hinstFreeTDS;
+#endif
+
+static void
+assert_equal_str(DSTR a, const char *b)
+{
+  assert(b && strncmp(tds_dstr_cstr(&a), b, tds_dstr_len(&a))==0);
```
fziglio, line 13: do you really want to test the beginning of the string or the entire one?

include/freetds/odbc.h
```diff
@@ -529,10 +529,7 @@ enum {
  * connectparams.h
  */
 
-typedef struct {
-	const char *p;
-	size_t len;
-} TDS_PARSED_PARAM;
+typedef DSTR TDS_PARSED_PARAM;
```
fziglio, line 532: I think it would be easier to pass `char*` instead of `const char*` and use the passed string (changing it), all callers uses copies of the original connection strings.

src/odbc/unittests/connection_string_parse.c
OverOrion: I'm not sure what you mean by this, could you please elaborate?

OverOrion:
Thank you for your review @fziglio, I've created fixup commits that hopefully addresses your notes, please let me know if I missed any or you have further notes!

src/odbc/unittests/connection_string_parse.c
fziglio: strncmp is comparing the beginning of the string. So `strncmp("foo", "foo123", 4)` is detecting the strings are the same

src/odbc/unittests/connection_string_parse.c
```diff
@@ -0,0 +1,198 @@
...
+}
+
+static int
+test_special_symbols(void)
+{
+  TDSLOGIN *login;
+  TDS_ERRS errs = {0};
+  TDSLOCALE *locale;
+  TDS_PARSED_PARAM parsed_params[ODBC_PARAM_SIZE];
+  const char *connect_string = "DRIVER={libtdsodbc.so};SERVER={127.0.0.1};PORT={1337};UID={test_username};PWD={[]{}(),;?*=!@};DATABASE={test_db};ClientCharset={UTF-8};";
```
fziglio, line 87: I tested with native Microsoft drivers. The `}` need to be quoted so if you want `PWD` to be `[]{}(),;?*=!@` the syntax is `PWD={[]{}}(),;?*=!@}` otherwise with `PWD={[]{}(),;?*=!@}` you will get `Invalid connection string attribute` error

fziglio:
Hi, sorry for the delay. After some though I was thinking that the change is large but more for the path that was taken than the real behaviour change. I tried to keep the test and I did a smaller change. What about last commits on https://github.com/FreeTDS/freetds/commits/appveyor ? Mainly I changed the way values are extracted keeping all the rest. Test passes and have no errors or leaks.

OverOrion:
Hi, terribly sorry for the delay. 
Thank you for your modification, it is greatly appreciated.
Added unit tests on top of your modifications.

PR #463, opened 2022-05-24
Add constants for table-valued parameters

brandonfoong:


include/sybdb.h
```diff
@@ -224,6 +224,8 @@ enum
 #define SYBMSDATETIME2 SYBMSDATETIME2
 	SYBMSDATETIMEOFFSET = 43, /* 0x2B */
 #define SYBMSDATETIMEOFFSET SYBMSDATETIMEOFFSET
+	SYBTABLETYPE = 243, /* 0xF3 */
+#define SYBTABLETYPE SYBTABLETYPE
```
fziglio, line 228: How do you plan to extend db-library to support this type? Better to avoid this definition

brandonfoong: Sorry about that, this PR was made by accident.

I'm hoping to add support for sending table-valued parameters. With reference to [pytds' implementation of TVPs](https://github.com/denisenkom/pytds/blob/7c73812444dd3c38c893dd88d14e138908069bd0/src/pytds/tds_types.py#L2245), I believe that we can use the existing serialization functions to send the table entries, and the implementation would mainly involve finding a suitable representation to store the TVP. Would greatly appreciate any advice you have on this matter!

PR #468, opened 2022-07-08
odbc: address comments on table-valued param binding

brandonfoong:
Follow up to #465.

https://github.com/FreeTDS/freetds/pull/465#discussion_r901062327: Newer bindings will overwrite the previous binding. For example, we bind the values [1, 2, 3] to an int column, and then bind [3, 4, 5] to the same column. When we send the table with `SQLExecDirect`, it will send [3, 4, 5]. I've updated `odbc_convert_table` to reflect this.

https://github.com/FreeTDS/freetds/pull/465#discussion_r901062701: Renamed `odbc_convert_table_row` to `odbc_convert_table_elem`, since the function adds a single element to the row.

https://github.com/FreeTDS/freetds/pull/465#discussion_r901062430: May I know what do you mean by "reset resetting the bindings"? From my understanding, if we have multiple table-valued parameters, setting `SQL_SOPT_SS_PARAM_FOCUS` will allow us to select which table we want to bind a column to. Also, before calling `SQLExecDirect`, we have to reset the param focus to 0.

https://github.com/FreeTDS/freetds/pull/465#discussion_r901062779: I believe that the CI checks already run the tests on the MS ODBC driver, is there anything else you'd like me to look out for?

### DIFF
--- a/src/odbc/sql2tds.c
+++ b/src/odbc/sql2tds.c
@@ -150,7 +150,7 @@ _odbc_blob_free(TDSCOLUMN *col)
 }
 
 static TDS_INT
-odbc_convert_table_row(TDS_DESC *apd, TDS_DESC *ipd,
+odbc_convert_table_elem(TDS_DESC *apd, TDS_DESC *ipd,
 	SQLUSMALLINT ipar, SQLSMALLINT fParamType, SQLSMALLINT fCType, SQLSMALLINT fSqlType,
 	SQLULEN cbColDef, SQLSMALLINT ibScale, SQLPOINTER rgbValue, SQLLEN cbValueMax, SQLLEN FAR *pcbValue)
 {
@@ -197,7 +197,7 @@ static TDS_INT
 odbc_convert_table(TDS_STMT *stmt, SQLTVP *src, TDS_TVP *dest, SQLLEN num_rows)
 {
 	SQLLEN i;
-	int j;
+	int j, res;
 	TDS_TVP_ROW *row;
 	TDS_TVP_ROW **prow;
 	TDSPARAMINFO *params, *new_params;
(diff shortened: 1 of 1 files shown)


src/odbc/sql2tds.c
```diff
@@ -237,11 +237,18 @@ odbc_convert_table(TDS_STMT *stmt, SQLTVP *src, TDS_TVP *dest, SQLLEN num_rows)
 	}
 
 	/* Convert the columns from a linked list to an array */
-	if ((cols = malloc(src->num_cols * sizeof(SQLTVPCOLUMN *))) == NULL)
+	if ((cols = calloc(src->num_cols, sizeof(SQLTVPCOLUMN *))) == NULL)
```
fziglio, line 240: You can also use the `tds_new0` utility. like `if ((cols = tds_new0(SQLTVPCOLUMN, src->num_cols)) == NULL)`

src/odbc/sql2tds.c
```diff
@@ -256,9 +263,11 @@ odbc_convert_table(TDS_STMT *stmt, SQLTVP *src, TDS_TVP *dest, SQLLEN num_rows)
 		if (!(new_params = tds_alloc_param_result(params)))
 			return TDS_CONVERT_NOMEM;
 
-		odbc_convert_table_row(apd, ipd, j + 1, cols[j]->fParamType, cols[j]->fCType,
+		res = odbc_convert_table_elem(apd, ipd, j + 1, cols[j]->fParamType, cols[j]->fCType,
 			cols[j]->fSqlType, cols[j]->cbColDef, cols[j]->ibScale,
 			NULL, cols[j]->cbValueMax, (SQLLEN *) SQL_NULL_DATA);
+		if (res != TDS_SUCCESS)
+			return res;
```
fziglio, line 270: Apparently you used 2 types of enumerations here. Conversion errors and TDS errors.

src/odbc/sql2tds.c
```diff
@@ -150,7 +150,7 @@ _odbc_blob_free(TDSCOLUMN *col)
 }
 
 static TDS_INT
-odbc_convert_table_row(TDS_DESC *apd, TDS_DESC *ipd,
+odbc_convert_table_elem(TDS_DESC *apd, TDS_DESC *ipd,
```
fziglio, line 153: I think `odbc_convert_table_field` would be less generic.

fziglio (commented):
Sorry for the delay, pretty busy period

fziglio:
> Follow up to #465.
> 
> [#465 (comment)](https://github.com/FreeTDS/freetds/pull/465#discussion_r901062327): Newer bindings will overwrite the previous binding. For example, we bind the values [1, 2, 3] to an int column, and then bind [3, 4, 5] to the same column. When we send the table with `SQLExecDirect`, it will send [3, 4, 5]. I've updated `odbc_convert_table` to reflect this.
> 

The issue is also that if you bind 1000 times you get 1000 allocations instead of 1.

> [#465 (comment)](https://github.com/FreeTDS/freetds/pull/465#discussion_r901062701): Renamed `odbc_convert_table_row` to `odbc_convert_table_elem`, since the function adds a single element to the row.
> 

Good, I think `odbc_convert_table_field` is even less generic.

> [#465 (comment)](https://github.com/FreeTDS/freetds/pull/465#discussion_r901062430): May I know what do you mean by "reset resetting the bindings"? From my understanding, if we have multiple table-valued parameters, setting `SQL_SOPT_SS_PARAM_FOCUS` will allow us to select which table we want to bind a column to. Also, before calling `SQLExecDirect`, we have to reset the param focus to 0.
> 

With `SQLFreeStmt(hstmt, SQL_RESET_PARAMS)` you can reset the parameters (also called bindings). Does this clear the "focus" too?

> [#465 (comment)](https://github.com/FreeTDS/freetds/pull/465#discussion_r901062779): I believe that the CI checks already run the tests on the MS ODBC driver, is there anything else you'd like me to look out for?

As far as I remember the CI checks with FreeTDS driver only.

freddy77:
Obsoleted, all issues resolved.

PR #474, opened 2022-12-07
Fix: heap-use-after-free reported by Address Sanitizer.

jbulow:
This fix prevents freetds to segfault when connection is lost to a SQLServer instance.

### DIFF
--- a/src/tds/net.c
+++ b/src/tds/net.c
@@ -749,7 +749,6 @@ tds_socket_read(TDSCONNECTION * conn, TDSSOCKET *tds, unsigned char *buf, int bu
 		return 0;
 
 	/* detect connection close */
-	tds_connection_close(conn);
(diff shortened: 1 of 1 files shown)


fziglio:
Sorry, this patch does not make much sense to me. The code was using `conn` already so you should get the memory issue before this function call. Can you describing what you were doing?

fziglio:
The issue could be the call to `tds_ssl_deinit` inside `tds_connection_close`.

jbulow:
And yes, removing tds_ssl_deinit inside tds_connection_close also solves the problem.

freddy77:
Fixed. New 1.3.16 version will be packaged automatically tomorrow. I also updated my settings to catch this type of errors (silly mistake).
Thanks.

PR #507, opened 2023-10-12
dblib: allow encryption option

coffeegreg:
The goal is to allow access to the `encryption` configuration option (`freetds.conf` setting) without the need for a `freetds.conf` file itself. I moved `DBSETENCRYPT` handling from `dbsetlbool` to `dbsetlname` (`dblib.c`) to handle `off/request/require` values and added `tds_set_encryption(TDSLOGIN * tds_login, const char *encryption)` to `login.c`.

### DIFF
--- a/include/sybdb.h
+++ b/include/sybdb.h
@@ -1220,7 +1220,6 @@ DBINT dbvarylen(DBPROCESS * dbproc, int column);
 #define SYBEBBFL        20297	/* -001- Batch failed in bulk-copy to SQL Server */
 #define SYBEDCL         20298	/* -004- DCL Error */
 #define SYBECS          20299	/* -004- cs context Error */
-#define SYBEINITLOGIN   20300	/* Initialization of the login structure failed */
 #define SYBEBULKINSERT  20599	/* cannot build bulk insert statement */
 #define SYBECOLSIZE     22000   /* Invalid column information structure size */
 
--- a/include/sybdb.h
+++ b/include/sybdb.h
@@ -1220,6 +1220,7 @@ DBINT dbvarylen(DBPROCESS * dbproc, int column);
 #define SYBEBBFL        20297	/* -001- Batch failed in bulk-copy to SQL Server */
 #define SYBEDCL         20298	/* -004- DCL Error */
 #define SYBECS          20299	/* -004- cs context Error */
+#define SYBEINITLOGIN   20300	/* Initialization of the login structure failed */
 #define SYBEBULKINSERT  20599	/* cannot build bulk insert statement */
 #define SYBECOLSIZE     22000   /* Invalid column information structure size */
 
--- a/include/sybdb.h
+++ b/include/sybdb.h
@@ -1278,7 +1278,7 @@ RETCODE dbsetlversion (LOGINREC * login, BYTE version);
 #define DBSETLPACKET(x,y)	dbsetllong((x), (y), DBSETPACKET)
 #define dbsetlpacket(x,y)	dbsetllong((x), (y), DBSETPACKET)
 #define DBSETENCRYPT		12
-#define DBSETLENCRYPT(x,y)	dbsetlname((x), (y), DBSETENCRYPT)
+#define DBSETLENCRYPT(x,y)	dbsetlbool((x), (y), DBSETENCRYPT)
 #define DBSETLABELED		13
 #define DBSETLLABELED(x,y)	dbsetlbool((x), (y), DBSETLABELED)
 #define BCP_SETLABELED(x,y)	dbsetlbool((x), (y), DBSETLABELED)
@@ -1300,6 +1300,8 @@ RETCODE dbsetlversion (LOGINREC * login, BYTE version);
 #define DBSETLREADONLY(x,y)	dbsetlbool((x), (y), DBSETREADONLY)
 #define DBSETDELEGATION		1004
 #define DBSETLDELEGATION(x, y)	dbsetlbool((x), (y), DBSETDELEGATION)
+#define DBSETENCRYPTION		1005
+#define DBSETLENCRYPTION(x, y)  dbsetlname((x), (y), DBSETENCRYPTION)
 
 RETCODE bcp_init(DBPROCESS * dbproc, const char *tblname, const char *hfile, const char *errfile, int direction);
 DBINT bcp_done(DBPROCESS * dbproc);
--- a/include/sybdb.h
+++ b/include/sybdb.h
@@ -1278,7 +1278,7 @@ RETCODE dbsetlversion (LOGINREC * login, BYTE version);
 #define DBSETLPACKET(x,y)	dbsetllong((x), (y), DBSETPACKET)
 #define dbsetlpacket(x,y)	dbsetllong((x), (y), DBSETPACKET)
 #define DBSETENCRYPT		12
-#define DBSETLENCRYPT(x,y)	dbsetlbool((x), (y), DBSETENCRYPT)
+#define DBSETLENCRYPT(x,y)	dbsetlname((x), (y), DBSETENCRYPT)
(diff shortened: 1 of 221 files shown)


include/sybdb.h
```diff
@@ -1278,7 +1278,7 @@ RETCODE dbsetlversion (LOGINREC * login, BYTE version);
 #define DBSETLPACKET(x,y)	dbsetllong((x), (y), DBSETPACKET)
 #define dbsetlpacket(x,y)	dbsetllong((x), (y), DBSETPACKET)
 #define DBSETENCRYPT		12
-#define DBSETLENCRYPT(x,y)	dbsetlbool((x), (y), DBSETENCRYPT)
+#define DBSETLENCRYPT(x,y)	dbsetlname((x), (y), DBSETENCRYPT)
```
freddy77, line 1281: this breaks both ABI and API compatibility

src/tds/login.c
```diff
@@ -170,6 +170,22 @@ tds_set_language(TDSLOGIN * tds_login, const char *language)
 	return !!tds_dstr_copy(&tds_login->language, language);
 }
 
+bool
+tds_set_encryption(TDSLOGIN * tds_login, const char *encryption)
+{
+	if (stricmp(encryption,TDS_STR_ENCRYPTION_OFF)==0)
```
freddy77, line 176: Use `strcasecmp`, not `stricmp` for portability. Also there's a very similar function in `src/tds/config.c`, `tds_config_encryption`. You can also reach it calling `tds_parse_conf_section`.

include/sybdb.h
coffeegreg: Any suggestions on how to deal with this problem? `DBSETENCRYPT` in `dbsetlbool` is unimplemented and marked as TODO.

src/tds/login.c
coffeegreg: Sure. Can this be fixed by:
```
	case DBSETENCRYPT:
		tds_parse_conf_section(TDS_STR_ENCRYPTION, value_nonull, login->tds_login);
		return SUCCEED;
		break;
```
or maybe leave this `tds_set_encryption` and replace `stricmp` with `strcasecmp` only as you suggested?

freddy77: Calling `tds_parse_conf_section` seems fine.

include/sybdb.h
freddy77: From the API prospective a line like `DBSETLENCRYPT(login, TRUE/FALSE)` should continue to compile. From the ABI prospective `dbsetlbool(login, TRUE/FALSE, 12)` should continue to set something for encryption. I would personally add a new setting, it could be something like
```c
#define DBSETENCRYPTION  1005
#define DBSETLENCRYPTION(x, y)  dbsetlname((x), (y), DBSETENCRYPTION)
```
`ENCRYPTION` is just a suggestion. The `1005` was the next FreeTDS extension from `include/sybdb.h`.

src/dblib/dblib.c
```diff
@@ -803,6 +803,10 @@ dbsetlname(LOGINREC * login, const char *value, int which)
 	case DBSETSERVERPRINCIPAL:
 		copy_ret = !!tds_dstr_copy(&login->tds_login->server_spn, value_nonull);
 		break;
+	case DBSETENCRYPTION:
+		tds_parse_conf_section(TDS_STR_ENCRYPTION, value_nonull, login->tds_login);
+		return SUCCEED;
```
freddy77, line 808: There's no way to return error from `tds_parse_conf_section` but maybe instead of
```c
return SUCCEED;
```
you could do something like
```c
copy_ret = !!login->tds_login->valid_configuration;
```
to detect error in login configuration.

coffeegreg: I'm not sure about this...
I've checked that regardless of the validity of the `encryption` parameter value `off/request/require` or any other invalid values `login->tds_login->valid_configuration` is `0` before calling `tds_parse_conf_section` and this value does not change after executing `tds_parse_conf_section`.
Let's take a look inside `tds_parse_conf_section` :
```
	} else if (!strcmp(option, TDS_STR_ENCRYPTION)) {
		tds_config_encryption(value, login);
	}
```
I've checked that `login->valid_configuration` is `0` before calling of `tds_config_encryption` and of course after execution of `tds_config_encryption` is still `0` because `tds_config_encryption` does not set `valid_configuration` to `1` anywhere.
Should we set `valid_configuration` to `1` before calling `tds_parse_conf_section` or set it with `tds_config_encryption`, or maybe do you have any other idea to deal with this problem?

freddy77: `valid_configuration` is set initially to 1 (true) and reset (set to false) if any error happens.

Yes, you are right, in this case if we would like to detect the error excluding previous ones we would need to set to 1 and check if it's set to 0 by `tds_parse_conf_section`.

coffeegreg: I thought that `valid_configuration` should be initially set to `1` but in fact it is not initialized or I can't find it.
I found comment in `dblogin` `/* set default values for loginrec */` but nothing here sets default values for `loginrec` at all. So I thought of calling `tds_init_login` here. Now, we have the `LOGINREC` structure initialized and of course also the `valid_configuration`.
However, all this does not change the fact that assigning the wrong value to `encryption` causes the connection to be established and the default value to be set `require`, assuming `require` is safer than `no`.

freddy77: I think I'm creating more confusion than needed. Reading the code and trying to remember now I understood. The `valid_configuration` is mainly used to detect wrong configuration from the file (like `freetds.conf`), not from user configuration. Time ago there were 2 structures, `TDS_LOGIN` and `TDS_CONNECTION`, first used to store the configuration from user, second to store configuration from file and final configuration (used for connection).
So a library usually:
- creates a `TDS_LOGIN`
- fills the values (like using `dbsetlname`)
- creates a `TDS_CONNECTION`
- fill the `TDS_CONNECTION` from file information (you need server name from `TDS_LOGIN`
- merge configuration from `TDS_LOGIN` to `TDS_CONNECTION`
- connect to server using `TDS_CONNECTION`

The `valid_configuration` was introduces for `TDS_CONNECTION` (now the 2 types points to the same structure), so not to detect errors from user settings (that's the reason code works with `valid_configuration` set to false on `TDS_LOGIN`).

I suggested (probably a wrong suggestion) to use `valid_configuration` (thinking that this flag was used by `TDS_LOGIN` also) to detect a single error for the specific setting but the code indeed is getting more complicated. Looking again at the code I think a more direct error not changing `valid_configuration` would be better.

freddy77: This should simplify the job https://github.com/FreeTDS/freetds/commit/e20a6515f58bdf4a27f937e069fda8bc6ca4116e (already in master)

coffeegreg: Great! This changes a lot and makes it easier.
Thank you for all the tips and information. Always good to know.

freddy77:
@coffeegreg what about https://github.com/FreeTDS/freetds/commit/18b31f5f917188e257efee5ddc0350e8aa77f283 (merged and updated commit message, kept author obviously). There's no `Signed-off-by`, should I add one?

coffeegreg:
Add whatever you need.
My changes were really cosmetic. You did the most work by changing tds_parse_conf_section and everything that use it.
Thank you.

freddy77:
> Add whatever you need. My changes were really cosmetic. You did the most work by changing tds_parse_conf_section and everything that use it. Thank you.

Fine. Do you want to provide a real name/email or should I leave coffeegreg <86416144+coffeegreg@users.noreply.github.com> ?

coffeegreg:
Leave with coffeegreg... :thumbsup:
Thanks again.

freddy77:
Merged, thanks

PR #524, opened 2024-02-20
datacopy: allow when dest_collen > src_collen

chenall:


freddy77:
It makes sense. Can you add some more description in the commit message (like the reason why this is fine) and `Signed-off-by` ?

chenall:
Sorry, the description may be not good.
Is it okay like this?

freddy77:
Merged

PR #527, opened 2024-03-04
add port method

taozuhong:


include/sybdb.h
```diff
@@ -1285,6 +1285,8 @@ RETCODE dbsetlversion (LOGINREC * login, BYTE version);
 #define DBSETDBNAME		14
 #define DBSETLDBNAME(x,y)	dbsetlname((x), (y), DBSETDBNAME)
 #define DBSETLVERSION(login, version) dbsetlversion((login), (version))
+#define DBSETPORT		15
```
freddy77, line 1288: If it does not come from Sybase/Microsoft should be put as an extension below.

src/dblib/dblib.c
```diff
@@ -854,7 +854,6 @@ dbsetllong(LOGINREC * login, long value, int which)
 	}
 }
 
-#if defined(DBLIB_UNIMPLEMENTED)
```
freddy77, line 857: Maybe also change the comment
```
 * \retval FAIL anything other than \c DBSETHIER was passed for \a which.  
```
to something more generic like
```
 * \retval FAIL if invalid field in \a which or invalid \a value.  
```

src/dblib/dblib.c
```diff
@@ -877,14 +876,17 @@ dbsetlshort(LOGINREC * login, int value, int which)
 	}
 
 	switch (which) {
-	case DBSETHIER:
+	case DBSETPORT:
+		tds_set_port(login->tds_login, value);
+		return SUCCEED;
+		break;
+	// case DBSETHIER:
```
freddy77, line 883: Do not use C++ style comments

taozuhong:
Merge all review suggestions into on commit (pls see #529)

PR #529, opened 2024-03-04
dblib: Add port setting

taozuhong:
Signed-off-by: [Tao zuhong](mailto:taozuhong@hotmail.com)


freddy77:
Code looks good. I would use the term "setting" instead of "method" in the commit, even better something like `dblib: Add port setting`. Can you also add `Signed-off-by` line? If you agree I can change and merge.

taozuhong:
> Code looks good. I would use the term "setting" instead of "method" in the commit, even better something like `dblib: Add port setting`. Can you also add `Signed-off-by` line? If you agree I can change and merge.

Done as your suggests.

freddy77:
Merged

Issue #528, opened 2024-03-04
tsql does not support non-standard ports in MS SQL Server.


chenall:
I have set up a trial MSSQLSERVER on port 2022, and there seems to be an issue with TSQL not connecting properly.

by the way, can it be modified to be compatible with isql and osql, support SERVER, PORT and so on?

e.g:
isql -Usa -Ppassword -S127.0.0.1,2022

![20240304133252](https://github.com/FreeTDS/freetds/assets/628011/6f7028e9-178c-4fbf-a6b7-def19d975216)

```
13:32:43.483 3248 (log.c:187):Starting log file for FreeTDS 1.5.dev.20240301
	on 2024-03-04 13:32:43 with debug flags 0xffff.
13:32:43.488 3248 (iconv.c:371):tds_iconv_open(023E6BB0, CP936, 1)
13:32:43.488 3248 (iconv.c:202):local name for ISO-8859-1 is ISO-8859-1
13:32:43.488 3248 (iconv.c:202):local name for UTF-8 is UTF-8
13:32:43.488 3248 (iconv.c:202):local name for UCS-2LE is UCS-2LE
13:32:43.488 3248 (iconv.c:202):local name for UCS-2BE is UCS-2BE
13:32:43.488 3248 (iconv.c:390):setting up conversions for client charset "CP936"
13:32:43.488 3248 (iconv.c:392):preparing iconv for "CP936" <-> "UCS-2LE" conversion
13:32:43.488 3248 (iconv.c:431):tds_iconv_open: done
13:32:43.488 3248 (net.c:392):Connecting with protocol version 7.4
13:32:43.488 3248 (net.c:318):Connecting to 127.0.0.1 port 1433
13:32:43.488 3248 (net.c:340):tds_setup_socket: connect(2) returned "无法立即完成一个非阻止性套接字操作。
"
13:32:45.490 3248 (net.c:231):getsockopt(2) reported: 由于目标计算机积极拒绝，无法连接。

13:32:45.490 3248 (net.c:527):tds_open_socket() failed
13:32:45.490 3248 (util.c:333):tdserror(0019FCC8, 023E6FE8, 20009, 10061)
13:32:45.490 3248 (util.c:363):tdserror: client library returned TDS_INT_CANCEL(2)
13:32:45.490 3248 (util.c:386):tdserror: returning TDS_INT_CANCEL(2)
13:32:45.495 3248 (util.c:333):tdserror(023E6B70, 023E6FE8, 20009, 10061)
[...]

PR #530, opened 2024-03-07
Compatible with isql/osql syntax 

chenall:
Allows using server`,`port to specify server address and port number #528

tsql -Usa -Ppassword -S127.0.0.1,2022

This change makes it easy to replace batch scripts written in isql/osql syntax by renaming tsql to isql/osql

### DIFF
--- a/src/tds/config.c
+++ b/src/tds/config.c
@@ -1306,10 +1306,14 @@ parse_server_name_for_port(TDSLOGIN * connection, TDSLOGIN * login, bool update_
 	/* IPv6 address can be quoted */
 	if (server[0] == '[') {
 		pSep = strstr(server, "]:");
+		if (!pSep)
+			pSep = strstr(server, "],");
 		if (pSep)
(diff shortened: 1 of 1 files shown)


freddy77:
Personally I would avoid it. Configuration is already confusing, adding weird settings does not help, the `server,port` should IMHO something to leave to ODBC.
Why don't you instead use `TDSPORT` override the port and unify the script (see https://www.freetds.org/userguide/envvar.html) ?

chenall:
It is not convenient to use TDSPORT because the port number is not fixed in different servers.

Modify the batch script to replace the original ~SERVER,PORT~ with `SERVER:PORT` can also be used.
I am lazy, so I think it is more convenient to modify the code directly for compatibility. ^_^

freddy77:
I'm still thinking not to make configuration more complicated.
Willing to help with the batch script, it should not be complicated to write some variable/function to simplify the script.

PR #537, opened 2024-03-18
add open connection direct support

taozuhong:
If the `server` is null, then open connection direct with login information.

### DIFF
--- a/include/sybdb.h
+++ b/include/sybdb.h
@@ -1302,7 +1302,7 @@ RETCODE dbsetlversion (LOGINREC * login, BYTE version);
 #define DBSETLDELEGATION(x, y)	dbsetlbool((x), (y), DBSETDELEGATION)
 #define DBSETENCRYPTION		1005
 #define DBSETLENCRYPTION(x, y)  dbsetlname((x), (y), DBSETENCRYPTION)
-#define DBSETPORT 		1006
+#define DBSETPORT 			1006
 #define DBSETLPORT(x,y) 	dbsetlshort((x), (y), DBSETPORT)
 
 RETCODE bcp_init(DBPROCESS * dbproc, const char *tblname, const char *hfile, const char *errfile, int direction);
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -1299,58 +1299,6 @@ tdsdbopen(LOGINREC * login, const char *server, int msdblib)
 		tds_fix_login(connection);		/* initialize from Environment variables */
 	}
 
-	if (true) {
-		struct addrinfo *addrs;
-		char tmp[128];
-
-		tdsdump_log(TDS_DBG_INFO1, "Final connection parameters:\n");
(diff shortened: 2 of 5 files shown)


include/sybdb.h
```diff
@@ -1302,7 +1302,7 @@ RETCODE dbsetlversion (LOGINREC * login, BYTE version);
 #define DBSETLDELEGATION(x, y)	dbsetlbool((x), (y), DBSETDELEGATION)
 #define DBSETENCRYPTION		1005
 #define DBSETLENCRYPTION(x, y)  dbsetlname((x), (y), DBSETENCRYPTION)
-#define DBSETPORT 		1006
+#define DBSETPORT 			1006
```
freddy77, line 1305: that's already indented correctly

src/dblib/dblib.c
```diff
@@ -1196,18 +1199,6 @@ tdsdbopen(LOGINREC * login, const char *server, int msdblib)
 		tdsdump_log(TDS_DBG_FUNC, "tdsdbopen(%p, %s, [%s])\n", login, server? server : "0x0", msdblib? "microsoft" : "sybase");
 	}
 
-	/*
-	 * Sybase supports the DSQUERY environment variable and falls back to "SYBASE" if server is NULL. 
-	 * Microsoft uses a NULL or "" server to indicate a local server.  
-	 * FIXME: support local server for win32.  
-	 */
-	if (!server && !msdblib) {
```
freddy77, line 1204: Did you read the comment before deleting it and code? This is the expected behaviour, you are not supposed to change it for compatibility reasons.

src/dblib/dblib.c
```diff
@@ -1258,13 +1241,115 @@ tdsdbopen(LOGINREC * login, const char *server, int msdblib)
 		return NULL;
 	}
 
-	connection = tds_read_config_info(dbproc->tds_socket, login->tds_login, g_dblib_ctx.tds_ctx->locale);
-	if (!connection) {
-		dbclose(dbproc);
-		return NULL;
+	// Open connection direct while the server is empty
```
freddy77, line 1244: I think you can get what you want passing a server as `hostname:port` instead

src/dblib/dblib.c
```diff
@@ -1258,13 +1241,115 @@ tdsdbopen(LOGINREC * login, const char *server, int msdblib)
...
+	if (NULL == server) {
+		/*
+		* Sybase supports the DSQUERY environment variable and falls back to "SYBASE" if server is NULL. 
+		* Microsoft uses a NULL or "" server to indicate a local server.  
+		* FIXME: support local server for win32.  
+		*/
```
freddy77, line 1250: this is the expected behaviour, you are changing it

taozuhong: I'm enhancing it, while the server is null and env is empty, then it will fail. 

But I have passed all connection args via LOGIN

src/dblib/dblib.c
taozuhong: it's a nice way to implement it.

PR #555, opened 2024-05-22
NCBI changes as of May 2024

ucko:
Hello again.

As noted in #219 (the remainder of which this PR supersedes), we've now advanced to 1.4.12.   I've proceeded to forward-port my changes to the trunk, and would appreciate it if you could please take a look and incorporate at least some of them. As usual, I'm happy to address any questions or concerns you might have.

Thanks!


freddy77:
That's quite some big series! Well... I suppose we can do it one at a time.

ucko:
> That's quite some big series! Well... I suppose we can do it one at a time.

Some of these commits stand on their own; with the help of git-explode (which has alas suffered some bit rot), I can substitute 30 smaller PRs, though the biggest would still have 15 commits and some others would have two or three.

freddy77:
If they are independent a simple cherry-pick is doable also. But probably splitting the series would be helpful. Maybe starting splitting the smaller ones and leave all the others in a single big series it's easier?

freddy77:
Can you add `Signed-off-by` lines?

ucko:
It looks like splitting out 1512645bc's changes to `bulk.c`, which are thankfully independent of its other changes, should let me split the largest series roughly in half.  OTOH, `git-explode` may have missed some purely logical dependencies for which I'll need to account explicitly.

ucko:
Withdrawing in favor of smaller PRs (#558-#589).

PR #556, opened 2024-05-23
datacopy: Increase CREATE TABLE command buffer to accommodate larger queries

john-kendall:
datacopy fails on tables with an unusually high number of columns.


freddy77:
Code update is pretty easy to understand. About the commit message is would improve a bit, something like `datacopy: Increase command buffer to accommodate larger queries` (plus signed off)

john-kendall:
Is there something else I should do? I've never used github.

freddy77:
Nothing, merged, thanks.

PR #558, opened 2024-05-30
Most BCP improvements from NCBI, as of May 2024

ucko:
This PR roughly corresponds to the primarily BCP-related commits from #555, adding 02aa63c per dependency considerations and omitting a couple of commits that can receive individual PRs, which I'll open shortly.

### DIFF
--- a/src/tds/bulk.c
+++ b/src/tds/bulk.c
@@ -36,6 +36,8 @@
 #include <stdlib.h>
 #endif /* HAVE_STDLIB_H */
 
+#include <ctype.h>
+
 #include <assert.h>
 
 #include <freetds/tds.h>
@@ -268,6 +270,26 @@ tds7_build_bulk_insert_stmt(TDSSOCKET * tds, TDSPBCB * clause, TDSCOLUMN * bcpco
 	return TDS_SUCCESS;
 }
 
+static const char * uc_str(const DSTR * ds)
+{
+	/* return strupper(strdup(tds_dstr_cstr(s))); */
+	const char * s = tds_dstr_cstr(ds);
+	size_t n = strcspn(s, "abcdefghijklmnopqrstuvwxyz");
+	if (s[n] == '\0') {
+		return s; /* already all uppercase */
+	} else {
+		char * result;
+		size_t i;
+		n = tds_dstr_len(ds);
+		result = malloc(n + 1);
+		for (i = 0; i < n; i++) {
+			result[i] = toupper(s[i]);
+		}
+		result[n] = '\0';
+		return result;
+	}
+}
+
 /**
  * Prepare the query to be sent to server to request BCP information
  * \tds
@@ -284,6 +306,7 @@ tds_bcp_start_insert_stmt(TDSSOCKET * tds, TDSBCPINFO * bcpinfo)
 		TDSCOLUMN *bcpcol;
 		TDSPBCB colclause;
 		char clause_buffer[4096] = { 0 };
+		bool triggers_checked = !bcpinfo->hint; /* Done on demand */
 
 		colclause.pb = clause_buffer;
 		colclause.cb = sizeof(clause_buffer);
@@ -300,8 +323,23 @@ tds_bcp_start_insert_stmt(TDSSOCKET * tds, TDSBCPINFO * bcpinfo)
 				continue;
 			if (!bcpinfo->identity_insert_on && bcpcol->column_identity)
 				continue;
-			if (bcpcol->column_computed)
-				continue;
+			if (bcpcol->column_computed) {
+				if ( !triggers_checked ) {
+					const char * uc_hint
+						= uc_str(&bcpinfo->hint);
+					if (strstr(uc_hint, "FIRE_TRIGGERS")) {
+						bcpinfo->with_triggers = true;
+					}
+					if (uc_hint
+					    != tds_dstr_cstr(&bcpinfo->hint)) {
+						free((char *)uc_hint);
+					}
+					triggers_checked = true;
+				}
+				if ( !bcpinfo->with_triggers ) {
+					continue;
+				}
+			}
 			tds7_build_bulk_insert_stmt(tds, &colclause, bcpcol, firstcol);
 			firstcol = 0;
 		}
@@ -365,7 +403,8 @@ tds7_send_record(TDSSOCKET *tds, TDSBCPINFO *bcpinfo,
 
 		if ((!bcpinfo->identity_insert_on && bindcol->column_identity) ||
 			bindcol->column_timestamp ||
-			bindcol->column_computed ||
+			(bindcol->column_computed
+			 && !bcpinfo->with_triggers) ||
 			!tds_bcp_is_bound(bcpinfo, bindcol)) {
 			continue;
 		}
@@ -880,7 +919,8 @@ tds7_bcp_send_colmetadata(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
 		bcpcol = bcpinfo->bindinfo->columns[i];
 		if ((!bcpinfo->identity_insert_on && bcpcol->column_identity) || 
 			bcpcol->column_timestamp ||
-			bcpcol->column_computed ||
+			(bcpcol->column_computed
+			 && !bcpinfo->with_triggers) ||
 			!tds_bcp_is_bound(bcpinfo, bcpcol)) {
 			continue;
 		}
@@ -902,7 +942,8 @@ tds7_bcp_send_colmetadata(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
 
 		if ((!bcpinfo->identity_insert_on && bcpcol->column_identity) || 
 			bcpcol->column_timestamp ||
-			bcpcol->column_computed ||
+			(bcpcol->column_computed
+			 && !bcpinfo->with_triggers) ||
 			!tds_bcp_is_bound(bcpinfo, bcpcol)) {
 			continue;
 		}
--- a/src/tds/bulk.c
+++ b/src/tds/bulk.c
@@ -74,6 +74,9 @@ static int tds5_bcp_add_variable_columns(TDSBCPINFO *bcpinfo, tds_bcp_get_col_da
 					 int offset, TDS_UCHAR *rowbuffer, int start, int *pncols);
 static void tds_bcp_row_free(TDSRESULTINFO* result, unsigned char *row);
 static TDSRET tds5_process_insert_bulk_reply(TDSSOCKET * tds, TDSBCPINFO *bcpinfo);
+static TDSRET tds5_get_col_data_or_dflt(tds_bcp_get_col_data get_col_data,
+					TDSBCPINFO *bulk, TDSCOLUMN *bcpcol,
+					int offset, int colnum);
 
 /**
  * Initialize BCP information.
@@ -439,7 +442,29 @@ tds5_send_record(TDSSOCKET *tds, TDSBCPINFO *bcpinfo,
 	for (i = 0; i < bcpinfo->bindinfo->num_cols; i++) {
 		TDSCOLUMN  *bindcol = bcpinfo->bindinfo->columns[i];
 		if (is_blob_type(bindcol->on_server.column_type)) {
-			TDS_PROPAGATE(get_col_data(bcpinfo, bindcol, offset));
+			/* Elide trailing NULLs */
+			if (bindcol->bcp_column_data->is_null) {
+				int j;
+				for (j = i + 1;
+				     j < bcpinfo->bindinfo->num_cols; ++j) {
+					TDSCOLUMN *bindcol2
+						= bcpinfo->bindinfo
+						->columns[j];
+					if (is_blob_type(bindcol2->column_type)
+					    &&  !(bindcol2->bcp_column_data
+						  ->is_null)) {
+						break;
+					}
+				}
+				if (j == bcpinfo->bindinfo->num_cols) {
+					i = j;
+					break;
+				}
+			}
+
+			TDS_PROPAGATE(tds5_get_col_data_or_dflt
+				      (get_col_data, bcpinfo, bindcol, offset,
+				       i));
 			/* unknown but zero */
 			tds_put_smallint(tds, 0);
 			TDS_PUT_BYTE(tds, bindcol->on_server.column_type);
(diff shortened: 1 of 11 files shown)


ucko:
I've force-pushed a corrected version superseding #570.

ucko:
Meanwhile, I'm looking to extend this series with BCP-specific tuneups superseding #573.  To that end, I wanted to compare behavior for three cases: normal BCP in, BCP into an alternate version of the table where column types are all `VARCHAR(64)` (to confirm that conversion to `VARCHAR(n)` is OK when truncation definitely won't be needed), and BCP into an alternate version where they're all `VARCHAR(1)` (to see what happens when truncation definitely **will** be needed), using `ctlib/unittests/blk_in.c` and `dblib/unittests/bcp.c` as starting points.

On the DBLIB front, I found that all three cases fully succeeded when built against this branch, but `VARCHAR(1)` failed for at least one column (details TBD) when using proprietary libraries, so more strictness may actually be in order.  On the CTLIB front, something appears to be off altogether, as even the initial run with the usual column types fails with
```
Open Client Message: 0x608010 0x6259b0
number 0x1010104 layer 1 origin 1 severity 1 number 4
msgstring: blk_describe(): blk layer: user error: Parameter colnum has an illegal value of 1
osstring: (null)
blk_describe(1) failed
Open Client Message: 0x608010 0x6259b0
number 0x1010104 layer 1 origin 1 severity 1 number 4
msgstring: blk_describe(): blk layer: user error: Parameter colnum has an illegal value of 2
osstring: (null)
blk_describe(2) failed
[...]
```
Any idea what I might be missing?  I took care to predefine `SYB_LP64` and verified that only FreeTDS-specific headers come from FreeTDS.

ucko:
Here are my summarized comparison results (comparing against version 16.0 of Sybase's client libraries), where `*B` is (`VAR`)`BINARY`; `*C` is (`VAR`)`CHAR`; `B`, `C`, `VB`, and `VC` are respectively `BINARY`, `CHAR`, `VARBINARY`, and `VARCHAR`; `*DT` is (`SMALL`)`DATETIME`; `*$` is (`SMALL`)`MONEY`; and `#` is any numerical type -- `BIT`, any variant of `INT`, `REAL`, `FLOAT`, `DECIMAL`, or `NUMERIC`:

## CTLIB BCP truncation differences:
```
FROM\TO	*B(1)	*B(64)	*C(1)	*C(64)
C(10)	a	a	b	OK
VC(10)	a	a	c	OK
B(10)	b	OK	c/d	OK
VB(10)	e	OK	e	OK
*DT	f/g	g	c	OK
*$/#	c/h	OK	e/h	OK
```
* a)
  * Sybase: clearly expects hex digits; behavior depends on what it sees first:
     * non-hex char among first 2n: rowxfer fails w/syntax error
     * excess char when length is implicit [C(n)]: xfer noisily truncates but OK(!)
     * excess char when length is given [VC(n)]: rowxfer fails w/overflow error
   * FreeTDS: rowxfer silently fails in all those cases
* b) silently truncates under Sybase (but OK), fails under FreeTDS
* c) noisily truncates in rowxfer under Sybase (but OK), fails under FreeTDS
* d) B(n) -> *C(m): silently truncates under Sybase when 4 <= m < 2n; the cutoff of 4 suggests a leading 0x, but VB(n) -> *C(2n) is OK
* e) rowxfer fails under both, (overflow) error only under Sybase
* f) S: bind fails (type mismatch), xfer OK (!); F: xfer fails; (g) if NULL
* g) rowxfer always OK, but Sybase bind fails (type mismatch)
* h) BIT is always OK, as is TINYINT -> *BINARY(1)

## DBLIB BCP truncation differences:
```
FROM\TO	*B(1)	*B(64)	*C(1)	*C(64)
*C(10)	a	a	b	OK
B(10)	c	OK	b/d	OK
VB(10)	b	OK	b/d	OK
*DT/*$	b	OK	b	OK
#	b/e	OK/f	b/e	OK/f
```

* a)
  * Sybase: syntax error if non-hex char among first 2n, otherwise overflow error as appropriate
  * FreeTDS: syntax error if non-hex char present at all, otherwise silently truncates
* b) conversion overflow error (20049) from Sybase, silent truncation by FreeTDS
* c) oversized row error (20074) from Sybase, silent truncation by FreeTDS
* d) *B(n) -> *C(m): silently truncates under Sybase when 4 <= m < 2n
* e) BIT is always OK, as is TINYINT -> *BINARY(1)
* f) Sybase crashes with heap corruption for DECIMAL and NUMERIC

ucko:
I'll turn back to other PRs for now, but do still plan to add a proper replacement for #573 at some point.

ucko:
Ah, at least on the FreeTDS side, some `ctlib` truncation messages have been coming from the CS layer, for which `blk_in` (like most `ctlib` unit tests) doesn't register a handler.  I'll resurvey with one in place.

ucko:
## CTLIB BCP truncation differences w/CS layer messages printed:
```
FROM\TO	*B(1)	*B(64)	*C(1)	*C(64)
C(10)	a	a	b	OK
VC(10)	a	a	c	OK
B(10)	b	OK	c/d	OK
VB(10)	e	OK	e	OK
*DT	f/g	g	c	OK
*$/#	c/h	OK	e/h	OK
```

* a)
  *  Sybase: clearly expects hex digits; behavior depends on what it sees first:
     * non-hex char among first 2n: rowxfer fails w/syntax error
     (1, 2, 1, 26) - `blk_rowxfer(): blk layer: internal BLK-Library error: Failed in conversion routine - syntax error.  col = # row = #.`
     * excess char when length is implicit [C(n)]: xfer noisily truncates but OK(!)
     (1, 2, 1, 42) - `blk_rowxfer(): blk layer: internal BLK-Library error: Data truncated while doing local character set conversion.  col = # row = #.`
     * excess char when length is given [VC(n)]: rowxfer fails w/overflow error
     (1, 2, 1, 25) - `blk_rowxfer(): blk layer: internal BLK-Library error: Failed in conversion routine - condition overflow.  col = # row = #.`
  * FreeTDS: expects hex digits with leading `0x`.
    * missing prefix, or non-hex char anywhere: rowxfer fails w/syntax error
     (2, 4, 1, 24) - `cs_convert: cslib user api layer: common library error: The conversion/operation was stopped due to a syntax error in the source field.`
    * excess char otherwise: rowxfer fails w/overflow error
     (2, 4, 1, 36) - `cs_convert: cslib user api layer: common library error: The result is truncated because the conversion/operation resulted in overflow.`
* b) silently truncates under Sybase (but OK), fails under FreeTDS -- (2, 4, 1, 36) as above
* c) noisily truncates in rowxfer under Sybase (but OK): (1, 2, 1, 42) as above
   fails under FreeTDS: (2, 4, 1, 36) as above
* d) B(n) -> *C(m): silently truncates under Sybase when 4 <= m < 2n; the cutoff of 4 suggests a leading 0x, but VB(n) -> *C(2n) is OK
* e) rowxfer overflow error - (1, 2, 1, 25) under Syb., (2, 4, 1, 36) under FTDS
* f)
  * Sybase: bind fails (type mismatch), blk_rowxfer OK (!)
     (1, 1, 1, 10) - `blk_bind(): blk layer: user error: a table column of type 1 cannot be bound to a program  variable of type 12.` (or `... 13.`)
  * FreeTDS: blk_rowxfer fails [(2, 4, 1, 36) as above]; (g) if NULL
* g) rowxfer always OK, but Sybase bind fails (type mismatch as above)
* h) BIT is always OK, as is TINYINT -> *BINARY(1)

Notably, all errors under Sybase officially come from the CT/BLK layer, whereas FreeTDS only ever issues CS-layer errors that offer less context and may be likelier to slip through the cracks.

ucko:
## Corrected DBLIB truncation differences  (modulo formatting details along the above lines and FreeTDS's support for additional client types):
```
FROM\TO	*B(1)	*B(64)	*C(1)	*C(64)
*C(10)	a	a	b	OK
*B(10)	b	OK	c	OK
*DT/*$	c	OK	c	OK
#	c/d	OK/e	c/d	OK/e
```

* a) Both expect hex digits with optional leading `0x` (and effectively prepend a `0` to the body if the length is odd), but their behavior when something is off otherwise varies.
  * Sybase: syntax error (20050) if non-hex char among first 2n(+2), otherwise overflow error (20049) as appropriate
  * FreeTDS: syntax error (20050) if non-hex char present at all, otherwise silently truncates
* b) oversized row error (20074) from Sybase, silent truncation by FreeTDS
* c) overflow error (20049) from Sybase, silent truncation by FreeTDS
* d) BIT is always OK, as is `TINYINT `-> `*BINARY(1)`
* e) Sybase crashes with heap corruption for `DECIMAL` and `NUMERIC`

ucko:
My latest version of this series adds a commit to address a corner case I recently encountered by (re)allowing BCP into computed columns in the context of a `FIRE_TRIGGERS` hint, since an `INSTEAD OF INSERT` trigger may well be in play.

freddy77:
Picking up a commit to replace one work I was doing, specifically https://github.com/FreeTDS/freetds/pull/558/commits/ae519f8edb55147f4d12f2c480baa0746900c4ae.

freddy77:
Another commit merged!

I'm rebasing the initial merge on master to get an updated version. See https://github.com/freddy77/freetds/tree/pr/558/ncbi-2024-05-bcp-improvements.

Some considerations
1. `_cs_convert` is getting a lot of different usages and parameters. The name starts to not be appropriate any more;
2. `_cs_convert` is returning a `CS_RETCODE` and also a `BLK_CONV_STATUS` which mostly looks like `tds_convert` result converted to a new enumeration. Wondering if there should be just a single return;
3. `bcp_column_data` is kind of weird. It stored TDS representation in the column. So why not usind `column_data` and other fields instead?
4. `bcp_column_data` character representation for Unicode data is always TDS raw data, so utf-16. For ctlib and dblib data is usually converted to client charset by libTDS, but not this data. Wondering if this inconsistency should be removed;
5. wondering how proprietary `bcp` tools export Unicode data. It looks like Sybase writes data as binary (so `'A'` -> `4100`), not sure about Microsoft;
6. should FreeTDS store format file (`freebcp` utility) with neutral data types (like `UNITEXT` for both MS and Sybase) ?

src/tds/bulk.c
```diff
@@ -258,6 +271,23 @@ tds7_build_bulk_insert_stmt(TDSSOCKET * tds, TDSPBCB * clause, TDSCOLUMN * bcpco
...
 }
 
+static const char * uc_str(const char * s)
+{
+	/* return strupper(strdup(s)); */
+	size_t n = strcspn(s, "abcdefghijklmnopqrstuvwxyz");
+	if (s[n] == '\0') {
+		return s; /* already all uppercase */
+	} else {
+		char * result = malloc(n + 1);
```
freddy77, line 281: Are you sure about this ? Isn't this function converting ASCII to upper case but stopping at first lower case character ?

src/tds/config.c
```diff
@@ -91,7 +91,7 @@ static tds_dir_char *interf_file = NULL;
 
 #define TDS_ISSPACE(c) isspace((unsigned char ) (c))
 
-const char STD_DATETIME_FMT[] = "%b %e %Y %I:%M%p";
+const char STD_DATETIME_FMT[] = "%Y-%m-%d %H:%M:%S.%z";
```
freddy77, line 94: Sure useful for BCP but I suppose it breaks a lot of tests

ucko:
> Another commit merged!
> 
> I'm rebasing the initial merge on master to get an updated version.

Thanks for all that and for taking another look in general, and sorry for not getting back to you sooner -- I work for a US federal agency, and the new administration imposed a broad communications ban until recently.

> 1. `_cs_convert` is getting a lot of different usages and parameters. The name starts to not be appropriate any more;

Is there some other name you'd prefer?  `_cs_transform`, perhaps?

> 2. `_cs_convert` is returning a `CS_RETCODE` and also a `BLK_CONV_STATUS` which mostly looks like `tds_convert` result converted to a new enumeration. Wondering if there should be just a single return;

Fair point.  I was also using the presence of a non-`NULL` bulk-conversion-status pointer as an indication that BCP semantics are in order, but I see that checking for a non-`NULL` handle should suffice; my latest version of this commit, which I've just force-pushed, adds a comment noting that only `_blk_get_col_data` is expected to supply one.  At any rate, I do see that `CS_FAIL` always accompanies both `BLK_CONV_OVERFLOW` and `..._SYNTAX_ERROR` and that `CS_SUCCEED` always accompanies `BLK_CONV_TRUNCATION`, so I went ahead with a single return value and adjusted `_blk_get_col_data` to treat `..._TRUNCATION` as an alternative successful return value.  This approach won't scale as well as what I previously had, but I suppose it doesn't need to.

> 3. `bcp_column_data` is kind of weird. It stored TDS representation in the column. So why not usind `column_data` and other fields instead?

Good question.  I was preserving the status quo here, but perhaps some simplification is possible.

> 4. `bcp_column_data` character representation for Unicode data is always TDS raw data, so utf-16. For ctlib and dblib data is usually converted to client charset by libTDS, but not this data. Wondering if this inconsistency should be removed;

Wouldn't that be a subtle API break?

> 5. wondering how proprietary `bcp` tools export Unicode data. It looks like Sybase writes data as binary (so `'A'` -> `4100`), not sure about Microsoft;

By my reading of https://learn.microsoft.com/en-us/sql/tools/bcp-utility?view=sql-server-ver16, it defaults to a legacy 8-bit encoding (e.g. CP-1252), but the Windows version can be directed to use UTF-16 instead.  I'm not sure we have any installations here I can use for empirical testing.  At any rate, I'm specifically concerned with bulk *insertion*. 

> 6. should FreeTDS store format file (`freebcp` utility) with neutral data types (like `UNITEXT` for both MS and Sybase) ?

Perhaps, but that's also out of scope here.

src/tds/bulk.c
ucko: Oops, yeah, I need to reset `n` to the length before proceeding here.

src/tds/config.c
ucko: Don't worry, I checked for regressions and didn't find any.

src/tds/bulk.c
ucko: Alternatively, the calling code could use `strcasestr`, but that's not portable, so we'd need to add an implementation to `replacements`.

src/tds/config.c
mmcnabb-vms: I found from profiling on VMS that a significant amount of time was spent in `strftime` for `bcp out -c` operations (actually more than 50% of the entire runtime), so I think there are noticeable improvements to be had by a hand-tuned date/time format for bcp operations . 

The ASE bcp client appears to use locale for `bcp out -c` date formatting but I don't think that's a requirement we need to follow

mmcnabb-vms:
@freddy77 I have found that all of the other commits from this PR can be applied in order , with a few trivial conflict resolutions, and the tests all pass against both MS (tdsver 7.4) and Sybase.  See [mm_pr558](https://github.com/FreeTDS/freetds/compare/master...mmcnabb-vms:freetds:mm_pr558?expand=1) for reference.  I haven't done much testing outside of the FreeTDS test suite though (e.g. the trigger application and blob input). 

This was just before your cherry-pick of "BCP into computed columns again..."

Are you planning to pick all of these  (just so I can rebase my #685 changes to the right place)

freddy77:
@mmcnabb-vms added a commit to replace entirely the "ctlib: Optimize blob conversion for bulk transfers in." commit in this series. `_cs_convert` was getting too messy and complicated. That's introduce quite some conflicts.

The series is surely nice but it lacks entirely tests. Not clear if it's doing what is supposed to do or not. For instance the "Fix DATE and DATETIME2 bulk insertion via TDS 7.0 and 7.1." is not fixing the issue.

freddy77:
Slowly getting this in. More than merging is a kind of rewriting. Now the behavior of char -> binary in ctlib is correct. Added tests for this.

PR #560, opened 2024-05-30
Cleanups from NCBI as of May 2024

ucko:
This PR contains three formally interdependent commits from #555; inasmuch as there's an overall theme, it's formal cleanups.


src/dblib/dbpivot.c
```diff
@@ -529,7 +530,7 @@ make_col_name(DBPROCESS *dbproc, const KEY_T *k)
 		return NULL;
 	}
 	for(pc=k->keys; pc < k->keys + k->nkeys; pc++) {
-		*s++ = strdup(string_value(pc));
+		*s++ = string_value(pc);
```
freddy77, line 533: nice spot

src/tds/challenge.c
```diff
@@ -174,13 +174,15 @@ make_ntlm_v2_hash(TDSSOCKET * tds, const char *passwd, unsigned char ntlm_v2_has
 	user_name = tds_dstr_cstr(&tds->login->user_name);
 
 	/* parse domain\username */
-	p = strchr(user_name, '\\');
+	p = user_name ? strchr(user_name, '\\') : NULL;
```
freddy77, line 177: this is useless, compiler bug?

src/tds/challenge.c
```diff
@@ -174,13 +174,15 @@ make_ntlm_v2_hash(TDSSOCKET * tds, const char *passwd, unsigned char ntlm_v2_has
...
 	domain = user_name;
-	domain_len = p - user_name;
+	domain_len = p ? p - user_name : 0;
 
-	user_name = p + 1;
-	user_name_len = strlen(user_name);
+	if (p != NULL) {
+		user_name = p + 1;
+		user_name_len = strlen(user_name);
+	}
```
freddy77, line 185: useless too, although in this case compiler can't much help, the username here has to have that kind of format

src/tds/config.c
```diff
@@ -700,7 +700,8 @@ tds_parse_conf_section(const char *option, const char *value, void *param)
 	}
 
 	if (!s || got_error) {
-		login->valid_configuration = 0;
+		if (login)
+			login->valid_configuration = 0;
```
freddy77, line 704: `login` is used dozens of times in this function, we would have crash much sooner than here

src/tds/query.c
```diff
@@ -145,8 +145,10 @@ tds_convert_string(TDSSOCKET * tds, TDSICONV * char_conv, const char *s, int len
 	/* allocate needed buffer (+1 is to exclude 0 case) */
 	ol = il * char_conv->to.charset.max_bytes_per_char / char_conv->from.charset.min_bytes_per_char + 1;
 	buf = tds_new(char, ol);
-	if (!buf)
+	if (!buf) {
+		*out_len = 0;
 		return NULL;
+	}
```
freddy77, line 151: caller should handle the error due to `NULL` and not use `out_len`... but won't hurt initializing it

src/tds/vstrbuild.c
```diff
@@ -156,7 +156,7 @@ tds_vstrbuild(char *buffer, int buflen, int *resultlen, const char *text, int te
 		case CALCPARAM:
 			switch (*text) {
 			case '!':
-				if (pnum <= tokcount) {
```
freddy77, line 159: nice spot

src/tds/challenge.c
```diff
@@ -225,6 +227,9 @@ make_lm_v2_response(const unsigned char ntlm_v2_hash[16],
 	int mac_len = 16 + client_data_len;
 	unsigned char *mac;
 
+	if (client_data == NULL	 ||  client_data_len <= 0)
+		return NULL;
+
```
freddy77, line 232: useless code

src/odbc/convert_tds2sql.c
```diff
@@ -69,8 +69,14 @@ odbc_convert_char(TDS_STMT * stmt, TDSCOLUMN * curcol, TDS_CHAR * src, TDS_UINT
 
 	/* FIXME MARS not correct cause is the global tds but stmt->tds can be NULL on SQLGetData */
 	TDSSOCKET *tds = stmt->dbc->tds_socket;
+	TDSICONV *conv;
 
-	TDSICONV *conv = curcol->char_conv;
+	if (curcol == NULL) {
+		odbc_errs_add(&stmt->errs, "HY013", NULL);
+		return SQL_NULL_DATA;
+	}
```
freddy77, line 77: `curcol` is never `NULL` here.

freddy77:
Merged part of "[Formally address static analyzer concerns.](https://github.com/FreeTDS/freetds/pull/560/commits/d6ce7b0758dd9620e22bfb8aa8a8ac573f42f499)".

About the `test_assert.h`: I suppose the reason is to be able to override `assert` macro using that header. Then why not using a different macro, kind of `TDS_ASSERT` or similar?

src/tds/challenge.c
ucko: Good point.

src/tds/challenge.c
ucko: I was concerned that the (indirect) caller might have supplied an unsuitably formatted username, but a closer look reveals that  `tds7_send_auth`  checks for a little-endian UCS-2 backslash before kicking off the call chain that leads to `make_ntlm_v2_hash`.

src/tds/config.c
ucko: The previous uses are all contingent on specific option values, so a (hypothetical!) call along the lines of
```C
tds_parse_conf_section("unimplemented", value, NULL);
```
could otherwise crash right there, but I don't know why the analyzer singled that one line out.

src/tds/challenge.c
ucko: Good point.  Perhaps more `_len` variables and arguments should be `size_t` or some other unsigned type for clarity, though.

ucko:
Thanks!  As for `test_assert.h`, our version serves two purposes: forcing direct calls to `assert` (otherwise left stock) to take effect for the current compilation unit even in release mode (by undefining `NDEBUG` before including system `assert.h`) and, on Windows, arranging to suppress dialog boxes from crashes of any sort.

freddy77:
> Thanks! As for `test_assert.h`, our version serves two purposes: forcing direct calls to `assert` (otherwise left stock) to take effect for the current compilation unit even in release mode (by undefining `NDEBUG` before including system `assert.h`) and, on Windows, arranging to suppress dialog boxes from crashes of any sort so that they don't stall automated test runs.

`NDEBUG` should be manually undefined in all tests. But it seems to me a good reason to have a different macro name then. It's confusing having a code using standard macro names but needed them to be no standard macros. Would something like this https://github.com/freddy77/freetds/commit/bc1e07fffb736b743d172ccf3d5fe04e730d9f8d make sense (beside the dumb commit message)?

freddy77:
You can call `_set_error_mode` and/or `_set_abort_behavior` to change `assert`/`abort` behaviour on Windows. How to make sure they are called before assert instead of using some weird MS internals (which won't work possibly using GCC or other compilers) that's another issue. One way would be to define `assert` and call a "empowered" `_assert` function which would do everything needed. Or define in `common.c` a `main` that calls a `test_main` to be define by each test. At this point I'm starting thinking that having some common generic test code to be reused by all components would be helpful. The code that reads the `PWD` file is duplicated quite a lot for instance.

ucko:
Good idea.  I'll want to make this library distinct from `libcommon` so that the unit tests for `replacements` and `utils` can use it too.  Also, those tests (and some others) all work offline, and as such shouldn't care whether the `PWD` file exists, so the test `main` shouldn't insist on finding it itself.  (This is not just a hypothetical situation -- the C++ Toolkit runs the online tests via a wrapper that confirms that they work properly when run against both Sybase and MS SQL servers by varying `TDSPWDFILE`.)

freddy77:
> Good idea. I'll want to make this library distinct from `libcommon` so that the unit tests for `replacements` and `utils` can use it too. Also, those tests (and some others) all work offline, and as such shouldn't care whether the `PWD` file exists, so the test `main` shouldn't insist on finding it itself. (This is not just a hypothetical situation -- the C++ Toolkit runs the online tests via a wrapper that confirms that they work properly when run against both Sybase and MS SQL servers by varying `TDSPWDFILE`.)

I suppose by `libcommon` you mean the various `libcommon` generated by `unittests` directory. A more detailed idea is to have this new common library in `utils/unittests` providing utilities for all `unittests` directories. The history about directory and dependencies was that mainly `odbc`, `ctlib` and `dblib` all depended on `tds` while `replacements` was only to define portable implementation of some system functions (like `strlcpy`) not available in all systems. But lot of features provided by `tds` started to be protocol/database independent and some started to be added to `replacements` even if there were not really replacements. So the `utils` was born, to separate common functions not dealing with database, protocol or strict portability. The `libcommon` in `unittests` instead were born to avoid compiling `common.c` and other files for each test program and also to make `CMake` less of a burden to maintain.

It's not clear from your paragraph the part about offline tests and online tests. One part of the paragraphs says that "those tests (and some others) all work offline" while another that "C++ Toolkit runs the online tests via a wrapper". So apparently one sentence applies to some tests and latter to other ones.

freddy77:
Merged [Fix compilation errors in single-threaded builds.](https://github.com/FreeTDS/freetds/pull/560/commits/e8f23eb82b1954ef2d8524f6f082a72a6755d451) and [Formally address static analyzer concerns.](https://github.com/FreeTDS/freetds/pull/560/commits/bbe2948d7514c5fe62666ed845e678fdfb4b0ac0)

include/freetds/utils/test_base.h
```diff
@@ -0,0 +1,64 @@
+#ifndef _tdsguard_afBM6E9n8CuIFSBHNNblq5
```
freddy77, line 1: I'll add copyright lines.

include/freetds/utils/test_base.h
```diff
@@ -0,0 +1,64 @@
...
+/*
+ * Base header for FreeTDS unit tests, even those just covering helpers
+ * from the utils and replacements trees.  Should be included first
+ * (possibly via a common.h) to be certain of preceding <assert.h>.
+ */
+
+/* Ensure assert is always active. */
+#undef NDEBUG
+#ifdef assert
+#  error "Include test_base.h (or common.h) earlier"
```
freddy77, line 13: This test turns out to be too strong for some tests. The problem is that some tests include, at the beginning, a C file and that's supposed to be the first thing they do (all some linking trick). Would be enough to check that NDEBUG is not defined instead ?

freddy77:
Updated and merged last bits.

PR #561, opened 2024-05-30
ct.c tuneups from NCBI as of May 2024

ucko:
This PR contains two tuneups to `ct.c` from #555 that `git-deps` flagged as interdependent, though it looks like that may have been a false positive.

### DIFF
--- a/src/ctlib/ct.c
+++ b/src/ctlib/ct.c
@@ -154,6 +154,10 @@ _ct_get_user_api_layer_error(int error)
 	case 15:
 		return "Use direction CS_BLK_IN or CS_BLK_OUT for a bulk copy operation.";
 		break;
+	case 36:
+		return "The result is truncated because the"
+		       " conversion/operation resulted in overflow.";
+		break;
 	case 51:
 		return "Exactly one of context and connection must be non-NULL.";
 		break;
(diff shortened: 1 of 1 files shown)


freddy77:
Merging first commit.

About the second, I'm doing some tests. Looks like the check on Sybase library is done while calling `ct_param`, not sending. Also the error is pretty different, I'm getting (`CS_CHAR_TYPE`):
```
Open Client Message: 0x2f658890 0x2f674180
number 0x101018a layer 1 origin 1 severity 1 number 138
msgstring: ct_param(): user api layer: external error: A data length of 25600 exceeds the maximum length allowed for CHAR data.
osstring: (null)
ct_param(char) failed
```
Also it looks like data for parameters are copied during ct_param (changing them after do not affect the data used in the query).

ucko:
Thanks!  I'll take a closer look at the second.

freddy77:
No rush. But the difference about `ct_param` getting the data is more concerning to me (well, not that affect this PR).
Consider a program that:
- computes value allocating memory;
- pass value to `ct_param`;
- free the value;
- calls `ct_send`.
At the time of `ct_send` if we attempt to get the data having just saved the pointer value could be garbage or released to OS, so potentially causing corruption (invalid data inserted) or crashes.

Anyway, OT.

PR #562, opened 2024-05-30
_blk_get_col_data: Consistently return TDS_FAIL on failure.

ucko:
Most failure scenarios accidentally returned CS_FAIL, which ironically corresponds to TDS_SUCCESS.

Split from #555.


freddy77:
Argh.... given also that the values are different and `CS_FAIL` is `TDS_SUCCESS` is not exactly a style change!

PR #563, opened 2024-05-30
Address "declared with greater visibility" warnings.

ucko:
Ensure that members of types that should be unhidden have unhidden forward declarations; likewise for their members, recursively.  To that end, give CS_SERVERMSG_*'s underlying struct/union types explicit names (cs_servermsg_*).

Split from #555.

### DIFF
--- a/include/ctlib.h
+++ b/include/ctlib.h
@@ -29,6 +29,18 @@
  * Internal (not part of the exposed API) prototypes and such.
  */
 
+/* Forward declarations to fix "declared with greater visibility" warnings */
+struct _cs_dynamic;
+struct _cs_param;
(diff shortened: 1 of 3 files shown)


freddy77:
I would say this patch is doing exactly the opposite of what is expected, these structure are meant to not be visible externally.

ucko:
Good point, a closer look revealed that I encountered such warnings only from outside projects that were using private FreeTDS headers.  I'll withdraw this PR accordingly and instead work on proposing public APIs for any needed functionality the CTLIB front-end doesn't already expose.

ucko:
FTR, #623, #624, and #625 should take care of my needs more cleanly; thanks for merging them!

PR #564, opened 2024-05-30
cspublic.h: Formally define BLK_VERSION_{155,157,160}

ucko:
... at least in the absence of CS_NO_LARGE_IDENTIFIERS; the corresponding CS_VERSION_* identifiers will otherwise be undefined anyway.

Split from #555.


ucko:
Thanks!

PR #566, opened 2024-05-30
tds_setup_connection: Skip nvc/uvc check under OpenServer.

ucko:
OpenServer instances may well report "Unknown language request" for it, resulting in outright login failures.

Split from #555.


freddy77:
Merged

PR #567, opened 2024-05-30
SQLCancel: Lock DBC mutex around extracting tds.

ucko:
Split from #555.


freddy77:
This smells like potential dead lock to me. In theory in Unix SQLCancel can be called inside a signal handler.

ucko:
Good point. In that case, I suppose the way to go would be to set some flag on the DBC itself and check it reasonably often from other functions (perhaps by extending the macros around entry and exit); pretty much anything beyond that (even a `trylock` call) would be unsafe to call from a signal handler. :-/

freddy77:
This last change looks overkilling. It reminds me of the hero movies in the 80-ish when the hero tried to pull out a bullet using a knife bigger than my forearm and trying not to hurt himself.
Joking aside it looks introducing more issues than what it's fixing. Surely needs more testing. But cancel have to do the cancel, not setting a flag and hoping for the best. Unfortunately too many cases are not tested, at least automatically. Should remove all data pending on same thread and statement? What the behaviour if connection is fully idle? What if the statement is not active? What if there are multiple active statement and you cancel one (MARS, obviously)?

ucko:
I hear you, and had initially looked into finding ways to make immediate cancellation safe.

Alas, there are severe limits on what code potentially called from a signal handler can safely do.  ISO is of course [particularly strict](https://en.cppreference.com/w/c/program/signal#Signal_handler); [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-170) is marginally better but still notably forbids any I/O.  [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_04) is more reasonable, but forbids any mutex operations, even `trylock`.  In practice, temporarily masking signals around lock operations might let `trylock` (and corresponding `unlock`) calls succeed safely, but would add major overhead, and AFAICT isn't even an option on Windows (which is too strict anyway).

Incidentally, I see from reviewing that documentation that the flag should have type `volatile sig_atomic_t` to be fully safe.

freddy77:
Some working on cancellation, adding new tests... and finding new bugs!

PR #568, opened 2024-05-30
Address Windows build errors, most config-dependent.

ucko:
* In the absence of an external SQLGetPrivateProfileString, fail the build only when expecting to use an external driver manager, and expose tds_SQLGetPrivateProfileString to the odbc "connect" unit test. Don't let that test attempt to use SQLGetPrivateProfileStringW even in Unicode builds.
* src/odbc/win*.c: Accommodate static builds, with no hinstFreeTDS; in particular, never attempt to show FreeTDS dialog boxes there.
* src/utils/unittests/challenge.c: #include <freetds/sysdep_private.h>
  to ensure the availability of strcasecmp.

Split from #555.

### DIFF
--- a/src/odbc/connectparams.c
+++ b/src/odbc/connectparams.c
@@ -95,8 +95,12 @@ static FILE *tdoGetIniFileName(void);
  *    expect see tdoGetIniFileName().
  *
  */
-static int SQLGetPrivateProfileString(LPCSTR pszSection, LPCSTR pszEntry, LPCSTR pszDefault, LPSTR pRetBuffer, int nRetBuffer,
-				      LPCSTR pszFileName);
+#ifndef _WIN32
+static
+#endif
+int SQLGetPrivateProfileString(LPCSTR pszSection, LPCSTR pszEntry,
+			       LPCSTR pszDefault, LPSTR pRetBuffer,
+			       int nRetBuffer, LPCSTR pszFileName);
 #endif
 
 #if defined(FILENAME_MAX) && FILENAME_MAX < 512
@@ -556,7 +560,7 @@ odbc_build_connect_string(TDS_ERRS *errs, TDS_PARSED_PARAM *params, char **out)
 
 #if !HAVE_SQLGETPRIVATEPROFILESTRING
 
-#ifdef _WIN32
+#if defined(_WIN32)  &&  !defined(TDS_NO_DM)
 #  error There is something wrong  in configuration...
 #endif
 
@@ -584,7 +588,10 @@ tdoParseProfile(const char *option, const char *value, void *param)
 	return true;
 }
 
-static int
+#ifndef _WIN32
+static
+#endif
+int
 SQLGetPrivateProfileString(LPCSTR pszSection, LPCSTR pszEntry, LPCSTR pszDefault, LPSTR pRetBuffer, int nRetBuffer,
 			   LPCSTR pszFileName)
 {
--- a/src/odbc/unittests/connect.c
+++ b/src/odbc/unittests/connect.c
@@ -12,18 +12,25 @@ init_connect(void)
 
 #ifdef _WIN32
 #include <odbcinst.h>
+#undef SQLGetPrivateProfileString
+#if !HAVE_SQLGETPRIVATEPROFILESTRING
+#  define SQLGetPrivateProfileString tds_SQLGetPrivateProfileString
+int tds_SQLGetPrivateProfileString(LPCSTR pszSection, LPCSTR pszEntry,
+                                   LPCSTR pszDefault, LPSTR pRetBuffer,
+                                   int nRetBuffer, LPCSTR pszFileName);
+#endif
 
 static char *entry = NULL;
 
 static char *
 get_entry(const char *key)
 {
-	static TCHAR buf[256];
-
+	static char buf[256];
+ 
 	entry = NULL;
-	if (SQLGetPrivateProfileString((LPCTSTR) T(odbc_server), (LPCTSTR) T(key), TEXT(""),
-				       buf, TDS_VECTOR_SIZE(buf), TEXT("odbc.ini")) > 0)
-		entry = C(buf);
+	if (SQLGetPrivateProfileString(odbc_server, key, "", buf,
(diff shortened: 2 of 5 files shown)


src/odbc/connectparams.c
```diff
@@ -556,7 +560,7 @@ odbc_build_connect_string(TDS_ERRS *errs, TDS_PARSED_PARAM *params, char **out)
 
 #if !HAVE_SQLGETPRIVATEPROFILESTRING
 
-#ifdef _WIN32
+#if defined(_WIN32)  &&  !defined(TDS_NO_DM)
```
freddy77, line 563: Personally I would avoid this, Windows has a standard driver manager which should be used

ucko:
I acknowledge that bypassing Windows' native DM is unconventional, but it's logistically simpler to use driverless mode everywhere than to arrange to link this API's tests directly against it only in the absence of a system DM.  (The C\+\+ Toolkit's build system covers bundled projects itself rather than using their own build systems.)  I also seem to recall that the system DM insisted on sticking to drivers that had been registered system-wide, but perhaps that's not actually a problem, at least nowadays.

freddy77:
Reasonable.

About the `connect.c` test... does it make sense to execute that part of the test that requires `get_entry` and so currently system `SQLGetPrivateProfileString`? It requires the driver to be registered as `FreeTDS` (see `DRIVER=FreeTDS;SERVER=%s;UID=%s;PWD=%s;DATABASE=%s;` connection string) which is only implemented by Driver Manager.

ucko:
AFAICT, that part of the test does actually succeed with this change in place, at least if I point the test at appropriate `freetds.conf` and `odbc.ini` files.  (I know there's logic to bypass that part when unable to look up the server name via `get_entry`, but it gets past that logic and proceeds with the remainder of that part.)

src/utils/unittests/challenge.c
```diff
@@ -35,6 +35,7 @@
 #endif
 
 #include "tds_sysdep_public.h"
+#include <freetds/sysdep_private.h>
```
freddy77, line 38: committed this hunk separately

src/odbc/winlogin.c
```diff
@@ -173,5 +173,9 @@ LoginDlgProc(HWND hDlg, UINT message, WPARAM wParam,	/* */
 bool
 get_login_info(HWND hwndParent, TDSLOGIN * login)
 {
+#ifdef DLL_EXPORT
```
freddy77, line 176: Why this is `DLL_EXPORT` and not, for instance `TDS_NO_DM` ?

src/odbc/winsetup.c
```diff
@@ -428,6 +433,7 @@ check_installer_errors(void)
 	return SQL_SUCCEEDED(ret);
 }
 
+#ifdef DLL_EXPORT
```
freddy77, line 436: yes, here it makes sense `DLL_EXPORT`, you cannot register something not an ODBC driver.

src/odbc/winlogin.c
freddy77: Okay, I got it, if you are not a DLL you don't have the dialog to load. But then all code above is not used too.

src/odbc/winsetup.c
```diff
@@ -318,7 +318,12 @@ ConfigDSN(HWND hwndParent, WORD fRequest, LPCSTR lpszDriver, LPCSTR lpszAttribut
 	/* Maybe allow the user to edit it */
 	if (hwndParent && fRequest != ODBC_REMOVE_DSN) {
 		INT_PTR result;
+#ifdef DLL_EXPORT
```
freddy77, line 321: I think all code in this file is exposing something for ODBC DM which cannot be used in case we are not a DLL.

src/odbc/unittests/connect.c
```diff
@@ -12,18 +12,25 @@ init_connect(void)
 
 #ifdef _WIN32
 #include <odbcinst.h>
+#undef SQLGetPrivateProfileString
+#if !HAVE_SQLGETPRIVATEPROFILESTRING
+#  define SQLGetPrivateProfileString tds_SQLGetPrivateProfileString
+int tds_SQLGetPrivateProfileString(LPCSTR pszSection, LPCSTR pszEntry,
+                                   LPCSTR pszDefault, LPSTR pRetBuffer,
+                                   int nRetBuffer, LPCSTR pszFileName);
```
freddy77, line 20: What happens if we use a DLL here? I would assume this test will refer to an undefined symbol and fail to load.

src/utils/unittests/challenge.c
ucko: Thanks!

src/odbc/winsetup.c
ucko: Further conditionalization may well be possible; I erred on the side of keeping more in.

src/odbc/unittests/connect.c
ucko: Ah, yeah, (conditional) exporting is in order here.  However, I need a static build anyway for the sake of the `all_types`, `connection_string_parse`, and `utf8_4` tests, so our Windows builds leave this library static-only for simplicity.

src/odbc/winsetup.c
freddy77: Merged the win* part.

src/odbc/unittests/connect.c
freddy77: Tried to sort out this last hunk. Maybe code for that specific function could be just duplicated somehow? I tried to build for Windows and `SQLGetPrivateProfileString` from odbc system libraries was used, not the version from shared library.

ucko:
I've added CMake tuneups I'd missed earlier and rebased against `master`.  With these tuneups in place, Windows builds still naturally default to the system driver manager, but I can build the `connect` test against `tds_SQLGetPrivateProfileString` by editing `config.h` to comment out `HAVE_SQLGETPRIVATEPROFILESTRING`'s definition and instead predefine `TDS_NO_DM`.

freddy77:
Merged

PR #569, opened 2024-05-30
_ct_get_client_type: Handle SYBUINTN.

ucko:
Split from #555.

### DIFF
--- a/src/ctlib/ct.c
+++ b/src/ctlib/ct.c
@@ -2142,6 +2142,20 @@ _ct_get_client_type(const TDSCOLUMN *col, bool describe)
 	case SYBUINT8:
 		return CS_UBIGINT_TYPE;
 		break;
+	case SYBUINTN:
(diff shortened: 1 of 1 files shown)


src/ctlib/ct.c
```diff
@@ -2142,6 +2142,20 @@ _ct_get_client_type(const TDSCOLUMN *col, bool describe)
 	case SYBUINT8:
 		return CS_UBIGINT_TYPE;
 		break;
+	case SYBUINTN:
```
freddy77, line 2145: `SYBUINTN` should be filtered out by `tds_get_conversion_type`, for the same reason `SYBINTN` is also not handled in this case.

ucko: Good point, thanks.  I'd carried this patch forward from an older version that dispatched directly on `tds_type` and handled `SYBINTN` but not `SYBUINTN`.

PR #570, opened 2024-05-30
tds_generic_put: Correct BCP 7.x new blob type handling

ucko:
... i.e., of VARCHAR(MAX) and the like.  Split from #555.

### DIFF
--- a/src/tds/data.c
+++ b/src/tds/data.c
@@ -1005,11 +1005,11 @@ tds_generic_put(TDSSOCKET * tds, TDSCOLUMN * curcol, int bcp7)
 
 		switch (curcol->column_varint_size) {
 		case 8:
-			/* this difference for BCP operation is due to
-			 * a bug in different server version that does
-			 * not accept a length here */
-			tds_put_int8(tds, bcp7 ? (TDS_INT8) -2 : (TDS_INT8) colsize);
-			tds_put_int(tds, colsize);
+			tds_put_int8(tds, colsize);
+			/* TDS 7.x no longer expects (or accepts) a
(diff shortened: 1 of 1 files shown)


freddy77:
That's... interesting! Tests passes with both this commit and without, so I suppose this path it's not really tested. Worth to write a test, and check TDS documentation!

ucko:
An analogous patch to 1.4 *has* been helpful for some of my usage.  However, when I locally extended `ctlib`'s `long_binary` test to incorporate BCP, I found that it succeeded only *without* this change.  I'll take a closer look and get back to you.

ucko:
AFAICT, whether this patch is helpful or harmful depends on runtime context TBD; I'll try to identify the relevant condition and adjust it accordingly.

freddy77:
TDS protocol documentation seems to indicate PLP length is needed. Yes, it would be useful to extend a test to catch this. I vaguely remember this specific code (I wrote it) and it was kind of a bug in a specific server version, or a combination of server and protocol version. Wondering why I didn't write a test at that time.

freddy77:
Added a test, see https://github.com/FreeTDS/freetds/commit/300611313f0e6d683bce436ac6b81df71265e4a2. Looks like it's working with current code, not with this patch applied.

freddy77:
Could it be that this commit was here to fix an old bug handling empty strings?

ucko:
Aha, the supporting logic for `blk_textxfer` in #558 doesn't account for possibly needing to send a final PLP length of zero.  I'll fix that instead.

ucko:
Withdrawn in favor of #558.

PR #571, opened 2024-05-30
tds5_process_dyn_result2: Don't trust Sybase w/blob params.

ucko:
As of ASE 16.0, Sybase servers have started allowing dynamic query (prepared statement) declarations with IMAGE or (N)TEXT parameters. However, subsequent attempts to instantiate these queries have been failing with message 3805, "The token datastream length was not correct."  In such cases, switch on dynamic query emulation (as already needed for older Sybase versions that immediately reject these declarations) and explicitly discard column information to avoid misconstruing the status of subsequent queries that yield no row results.

Split from #555.

### DIFF
--- a/src/tds/token.c
+++ b/src/tds/token.c
@@ -2687,6 +2687,7 @@ tds5_process_dyn_result2(TDSSOCKET * tds)
 	unsigned int col, num_cols;
 	TDSCOLUMN *curcol;
 	TDSPARAMINFO *info;
+	TDSDYNAMIC *dyn = NULL;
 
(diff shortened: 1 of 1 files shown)


freddy77:
I remember something like that, I tried to reproduce but I didn't manage. I tried with a `txt like ?` and a `insert into ( txt ) values ( ? )` and they worked (`ct_dynamic` using `CS_PREPARE` and `CS_EXECUTE`).

Do you have a reproduction query?

ucko:
> Do you have a reproduction query?

Not at present, sorry; my usual tests succeed at this point without this change.

PR #572, opened 2024-05-30
ct_data_info: Encode/decode CS_IODESC datatype field.

ucko:
Split from #555.


freddy77:
I think https://github.com/FreeTDS/freetds/commit/5f33559172c83972454270cc7c2a039d003168df took care of this issue.

ucko:
Thanks, but what about the other direction?

freddy77:
I look at the code and apparently, at the moment, it just stays there! Meaning you set into CT-Library and it's not used.
I suppose if someday it will be implemented we'll have to convert, if needed, to internal or server type.

freddy77:
Can I close this PR?

ucko:
OK, thanks.

PR #573, opened 2024-05-30
cs_convert, dbconvert: Truncate data as needed.

ucko:
* Automatically truncate strings that would overflow their destinations, mainly for the sake of bulk insertions.
* Fully honor *BINARY and IMAGE data buffer size limits in cs_convert.
* Update unit tests (ctlib cs_convert, dblib t0019) accordingly.

Split from #555.

### DIFF
--- a/src/ctlib/unittests/cs_convert.c
+++ b/src/ctlib/unittests/cs_convert.c
@@ -296,11 +296,13 @@ main(void)
 	DO_TEST(CS_INT test = 1234567;
 		CS_CHAR test2[] = "1234567", CS_INT_TYPE, &test, sizeof(test), CS_CHAR_TYPE, 7, CS_SUCCEED, test2, 7);
 	DO_TEST(CS_CHAR test[] = "abc";
-		CS_CHAR test2[] = "ab", CS_CHAR_TYPE, test, 3, CS_CHAR_TYPE, 2, CS_FAIL, test2, 2);
+		CS_CHAR test2[] = "ab", CS_CHAR_TYPE, test, 3, CS_CHAR_TYPE, 2,
+		CS_SUCCEED /* CS_FAIL */, test2, 2);
 	DO_TEST(CS_CHAR test[] = "abc";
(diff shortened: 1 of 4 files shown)


src/ctlib/unittests/cs_convert.c
```diff
@@ -296,11 +296,13 @@ main(void)
 	DO_TEST(CS_INT test = 1234567;
 		CS_CHAR test2[] = "1234567", CS_INT_TYPE, &test, sizeof(test), CS_CHAR_TYPE, 7, CS_SUCCEED, test2, 7);
 	DO_TEST(CS_CHAR test[] = "abc";
-		CS_CHAR test2[] = "ab", CS_CHAR_TYPE, test, 3, CS_CHAR_TYPE, 2, CS_FAIL, test2, 2);
+		CS_CHAR test2[] = "ab", CS_CHAR_TYPE, test, 3, CS_CHAR_TYPE, 2,
+		CS_SUCCEED /* CS_FAIL */, test2, 2);
```
freddy77, line 300: I think these tests are checked against proprietary libraries. That is you are introducing a bug.

ucko: Ah, yeah, looks like such truncation should occur only in the context of bulk insertion (where it *does* happen with proprietary libraries).

ucko:
Withdrawn in favor of a (planned) extension to #558.

ucko:
On second thought, reopening for completely reworked `dblib`-only changes (with `ctlib` changes going to #558):

    Match Sybase dblib's handling of overlong or malformatted BCP input.

    Adjust existing failure modes' boundaries and introduce the
    possibility of truncating with a message as a middle ground between
    silent truncation and noisy failure.  To that end, pass the
    destination's length to _bcp_convert_in.

ucko:
Actually reopening, oops.

ucko:
This extension wound up with only `ctlib` changes; I split the `dblib` changes off to #595.

PR #574, opened 2024-05-30
ct_connect: Set client_charset in addition to server_charset.

ucko:
Split from #555.


freddy77:
Not setting `client_charset` is done on purpose. This allows to pick up the server charset and convert properly on the client side.

ucko:
Good point.  FWIW, the original patch predated `CS_CLIENTCHARSET` support, and I hadn't registered the latter's availability.

PR #575, opened 2024-05-30
ct_describe: Issue an error for unsupported server types.

ucko:
Split from #555.


freddy77:
Wondering what Sybase would return in this case. Probably it should not happen in the sense that all types from Sybase server should be supported in some way.

ucko:
Right; IIRC, I was particularly concerned about `SQL_VARIANT`, which is MSSQL-only.

freddy77:
Merged

PR #577, opened 2024-05-30
dbwritetext: Fail if dbconvert does.

ucko:
Split from #555.

### DIFF
--- a/src/dblib/dblib.c
+++ b/src/dblib/dblib.c
@@ -6634,6 +6634,7 @@ RETCODE
 dbwritetext(DBPROCESS * dbproc, char *objname, DBBINARY * textptr, DBTINYINT textptrlen, DBBINARY * timestamp, DBBOOL log,
 	    DBINT size, BYTE * text)
 {
+	DBINT len = 0;
 	char textptr_string[35];	/* 16 * 2 + 2 (0x) + 1 */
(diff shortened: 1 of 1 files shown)


src/dblib/dblib.c
```diff
@@ -6652,8 +6653,14 @@ dbwritetext(DBPROCESS * dbproc, char *objname, DBBINARY * textptr, DBTINYINT tex
...
-	dbconvert(dbproc, SYBBINARY, (BYTE *) textptr, textptrlen, SYBCHAR, (BYTE *) textptr_string, -1);
-	dbconvert(dbproc, SYBBINARY, (BYTE *) timestamp, 8, SYBCHAR, (BYTE *) timestamp_string, -1);
+	len = dbconvert(dbproc, SYBBINARY, (BYTE *) textptr, textptrlen,
+			SYBCHAR, (BYTE *) textptr_string, -1);
+	if (len < 0)
+		return FAIL;
+	len = dbconvert(dbproc, SYBBINARY, (BYTE *) timestamp, 8, SYBCHAR,
+			(BYTE *) timestamp_string, -1);
+	if (len < 0)
+		return FAIL;
```
freddy77, line 6663: But this function cannot fail with these argument!

ucko:
Good point.  These checks come from a former colleague who left years ago; I'm unclear on why he added them.

PR #578, opened 2024-05-30
Acknowledge non-exhaustive TDS_SERVER_TYPE switch statements.

ucko:
Add no-op default cases to switch statements that don't explicitly cover all of TDS_SERVER_TYPE.


freddy77:
Mumble... this PR should be straightforward. Can I make the list explicit and add TODO comments, a bit like https://github.com/FreeTDS/freetds/blob/master/src/tds/query.c#L919 ?

ucko:
Sure, thanks!

freddy77:
Merged

PR #580, opened 2024-05-30
tds_iconv: Accommodate FreeBSD/Citrus iconv.

ucko:
Accommodate iconv implementations (such as the Citrus one used by modern FreeBSD systems) that ever indicate invalid input by reporting a positive count of irreversible characters rather than by setting errno to EILSEQ.

### DIFF
--- a/src/tds/iconv.c
+++ b/src/tds/iconv.c
@@ -657,6 +657,9 @@ tds_iconv(TDSSOCKET * tds, TDSICONV * conv, TDS_ICONV_DIRECTION io,
 
 		/* iconv success, return */
 		if (irreversible != (size_t) - 1) {
+			if (irreversible > 0) {
+				eilseq_raised = true;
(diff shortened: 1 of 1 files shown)


freddy77:
It looks like this change may ignore other library errors. Could we detect exactly of the implementation is really behaving like that?

freddy77:
I had a look at FreeBSD documentation and it looks like the behaviour is pretty POSIX compliant. Do you have an example of this issue?

ucko:
Yes, your `tds/unittests/utf8_2` has otherwise been failing here; it normally succeeds silently, but without the patch it instead exits with status 1 after printing
```
No warning returned in test 1 len 0
No warning returned in test 1 len 1
No warning returned in test 1 len 2
[...]
No warning returned in test 1 len 192
```

freddy77:
Merged

PR #581, opened 2024-05-30
Fix miscellaneous one-off warnings.

ucko:
* odbc.h: For Clang, substitute __attribute__((visibility("default"))) for __attribute__((externally_visible)), which it at least historically didn't support.
* odbc/unittests/common.h: When the ODBC 3 API is available, favor SQLSetStmtAttr over SQLSetStmtOption, which is deprecated.
* dblib/bcp.c (_bcp_get_col_data): Comment out an always-true comparison of column_bindlen (which was never signed) to zero.
* tds5_bcp_add_variable_columns: Explicitly mark 256 as unsigned to avoid spurious signed-vs.-unsigned comparison warnings.
* gssapi.c: On macOS, additionally suppress __API_DEPRECATED right before including gssapi_krb5.h (which is at least the last explicitly included system header).
* tls.c: Define _WIN32_WINNT ahead of config.h, which might otherwise supply a somewhat different definition, yielding a warning.
* ctlib/unittests/t0009.c: Pre-zero compute_col3.  Optimized builds with Clang or LLVM-based ICC can otherwise yield a Valgrind warning concerning a strcmp call that reads slightly past the trailing NUL.

### DIFF
--- a/src/dblib/bcp.c
+++ b/src/dblib/bcp.c
@@ -2226,7 +2226,7 @@ _bcp_get_col_data(TDSBCPINFO *bcpinfo, TDSCOLUMN *bindcol, int offset TDS_UNUSED
 
 	/* if (Max) column length specified take that into consideration. */
 
-	if (bindcol->column_bindlen >= 0) {
+	/* if (bindcol->column_bindlen >= 0) */ { /* bindlen is unsigned */
 		if (bindcol->column_bindlen == 0)
 			goto null_data;
 		if (collen)
--- a/src/odbc/unittests/rowset.c
+++ b/src/odbc/unittests/rowset.c
@@ -12,6 +12,20 @@ test_err(int n)
 	}
 }
 
+#ifdef _MSC_VER
+/* See https://learn.microsoft.com/en-us/cpp/preprocessor/warning?view=msvc-170 */
+#pragma warning(push)
+#pragma warning(disable:4996)
+static SQLRETURN
+_CHKSetStmtOption(SQLUSMALLINT option, SQLULEN value, const char* res)
+{
+    return CHKSetStmtOption(option, value, res);
+#pragma warning(pop)
+}
+#undef CHKSetStmtOption
+#define CHKSetStmtOption _CHKSetStmtOption
+#endif
+
 int
 main(void)
 {
--- a/src/tds/bulk.c
+++ b/src/tds/bulk.c
@@ -704,7 +704,7 @@ tds5_bcp_add_variable_columns(TDSBCPINFO *bcpinfo, tds_bcp_get_col_data get_col_
 
 	if (ncols) {
 		TDS_UCHAR *poff = rowbuffer + row_pos;
-		unsigned int pfx_top = offsets[ncols] / 256;
+		unsigned int pfx_top = offsets[ncols] >> 8;
 
 		tdsdump_log(TDS_DBG_FUNC, "ncols=%u poff=%p [%u]\n", ncols, poff, offsets[ncols]);
 
@@ -714,7 +714,7 @@ tds5_bcp_add_variable_columns(TDSBCPINFO *bcpinfo, tds_bcp_get_col_data get_col_
 			unsigned int n_pfx = 1;
 
 			for (i = 0; i <= ncols ; ++i)
-				if ((offsets[i] / 256) < pfx_top)
+				if ((offsets[i] >> 8) < pfx_top)
 					++n_pfx;
 			*poff++ = n_pfx;
 			--pfx_top;
--- a/src/tds/gssapi.c
+++ b/src/tds/gssapi.c
@@ -62,6 +62,8 @@
 #ifdef __APPLE__
 #define KERBEROS_APPLE_DEPRECATED(x)
 #define GSSKRB_APPLE_DEPRECATED(x)
+#undef __API_DEPRECATED
+#define __API_DEPRECATED(x, y)
 #endif
 #include <gssapi/gssapi_krb5.h>
 
@@ -238,8 +240,12 @@ tds_gss_get_auth(TDSSOCKET * tds)
 	 */
 	gss_buffer_desc send_tok;
 	OM_uint32 maj_stat, min_stat;
+#ifdef __APPLE__
+#  define nt_principal *(gss_OID_desc *) GSS_KRB5_NT_PRINCIPAL_NAME
+#else
 	/* same as GSS_KRB5_NT_PRINCIPAL_NAME but do not require .so library */
 	static gss_OID_desc nt_principal = { 10, (void*) "\x2a\x86\x48\x86\xf7\x12\x01\x02\x02\x01" };
+#endif
 	const char *server_name;
 	/* Storage for getaddrinfo calls */
 	struct addrinfo *addrs = NULL;
--- a/include/freetds/odbc.h
+++ b/include/freetds/odbc.h
@@ -74,7 +74,9 @@ extern "C"
 #endif
 
 #include <freetds/pushvis.h>
-#if defined(__GNUC__) && __GNUC__ >= 4 && !defined(__MINGW32__)
+#ifdef __clang__
+#define ODBC_API SQL_API __attribute__((visibility("default")))
+#elif defined(__GNUC__) && __GNUC__ >= 4 && !defined(__MINGW32__)
 #define ODBC_API SQL_API __attribute__((externally_visible))
 #else
 #define ODBC_API SQL_API
--- a/src/ctlib/unittests/t0009.c
+++ b/src/ctlib/unittests/t0009.c
@@ -37,7 +37,7 @@ main(void)
 	CS_CHAR col3[32];
 
 	CS_INT compute_col1;
-	CS_CHAR compute_col3[32];
+	CS_CHAR compute_col3[32] = "";
 
 	unsigned rows[3] = { 0, 0, 0 };
 


src/dblib/bcp.c
```diff
@@ -2226,7 +2226,7 @@ _bcp_get_col_data(TDSBCPINFO *bcpinfo, TDSCOLUMN *bindcol, int offset TDS_UNUSED
 
 	/* if (Max) column length specified take that into consideration. */
 
-	if (bindcol->column_bindlen >= 0) {
+	/* if (bindcol->column_bindlen >= 0) */ { /* bindlen is unsigned */
```
freddy77, line 2229: Not sure if this is what you want, or instead the variable should be signed instead. Depends on the usage, that variable is declared in libTDS but used in different ways by upper libraries. I need to check.

src/odbc/unittests/common.h
```diff
@@ -158,8 +158,13 @@ SQLSMALLINT odbc_alloc_handle_err_type(SQLSMALLINT type);
 	CHKR2(SQLSetPos, (odbc_stmt,a,b,c), SQL_HANDLE_STMT, odbc_stmt, res)
 #define CHKSetStmtAttr(a,b,c,res) \
 	CHKR2(SQLSetStmtAttr, (odbc_stmt,a,b,c), SQL_HANDLE_STMT, odbc_stmt, res)
+#if ODBCVER >= 0x0300
+#define CHKSetStmtOption(a,b,res) \
+	CHKSetStmtAttr(a, (SQLPOINTER) b, SQL_NTS, res)
+#else
```
freddy77, line 164: I would say no, the idea is exactly to check obsolete functions and that they are correct.

src/tds/bulk.c
```diff
@@ -714,7 +714,7 @@ tds5_bcp_add_variable_columns(TDSBCPINFO *bcpinfo, tds_bcp_get_col_data get_col_
 			unsigned int n_pfx = 1;
 
 			for (i = 0; i <= ncols ; ++i)
-				if ((offsets[i] / 256) < pfx_top)
+				if ((offsets[i] / 256u) < pfx_top)
```
freddy77, line 717: Maybe `>> 8` instead ? Similar code above.

src/tds/gssapi.c
```diff
@@ -238,8 +240,12 @@ tds_gss_get_auth(TDSSOCKET * tds)
 	 */
 	gss_buffer_desc send_tok;
 	OM_uint32 maj_stat, min_stat;
+#ifdef __APPLE__
+#  define nt_principal *(gss_OID_desc *) GSS_KRB5_NT_PRINCIPAL_NAME
+#else
```
freddy77, line 245: This is doing exactly the opposite of what the next line comment is suggesting we want.

src/tds/tls.c
```diff
@@ -18,11 +18,11 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#include <config.h>
-
```
freddy77, line 22: Why? This file is not supposed to include files so the `_WIN32_WINNT` definition won't change the result.

src/dblib/bcp.c
ucko: OK, thanks.  It's been unsigned all along, but I suppose that doesn't mean it can't change.

src/odbc/unittests/common.h
ucko: Windows builds of the `rowset` test otherwise warn
```
...\src\odbc\unittests\rowset.c(63): warning C4996: 'SQLSetStmtOption': ODBC API: SQLSetStmtOption is deprecated. Please use SQLSetStmtAttr instead.
```

src/tds/gssapi.c
ucko: Ah, sorry, I accidentally left the rationale out of my commit message, which should have included

> * tds_gss_get_auth: Use central GSS_KRB5_NT_PRINCIPAL_NAME on macOS.
>   It's available there with existing build settings, and avoids
> 
>     ld: warning: alignment (2) of atom '_tds_gss_get_auth_ver100.nt_principal'
>     is too small and may result in unaligned pointers 
>
>   which could otherwise crop up dozens of times.

IIRC, the warning occurs only when targeting x86 CPUs; Apple of course switched to "Silicon" ARM CPUs a couple of years ago, but still retains some support for x86.

src/tds/tls.c
ucko: Ah, right, this warning's specific to my setup -- in the course of integrating FreeTDS, I supply a `config.h` that sits on top of NCBI C++ Toolkit's configuration headers and as such indirectly pulls in a conditional `_WIN32_WINNT` definition.

src/odbc/unittests/common.h
ucko: I've reworked this commit to patch `rowset.c` rather than `common.h`  and have it still call `SetStmtOption`, just with the warning suppressed.

src/tds/gssapi.c
freddy77: It looks like a compiler/linker bug to me. Why the toolchain is not able to align it properly? It this a problem with data declared inside functions?

ucko: No, AFAICT the issue is that the relevant `struct` is underaligned on pre-Silicon macOS architectures because
```C
#if defined(__APPLE__) && (defined(__ppc__) || defined(__ppc64__) || defined(__i386__) || defined(__x86_64__))
#pragma pack(push,2)
#endif
```
is in effect for its definition.  I can force the compiler's hand by specifying something like
```C
	static struct {
		TDS_INT8 pad1;
		TDS_INT	 pad2;
		gss_OID_desc oid;
	} nt_principal =
		{ 0, 0,
		  { 10, (void*) "\x2a\x86\x48\x86\xf7\x12\x01\x02\x02\x01" } };

```
to ensure that the structure starts at a suitable phase (4 mod 8) and then referring to `nt_principal.oid`, but that's ugly and adds overhead, whereas my proposed change simply takes advantage of a stock definition that's available anyway.  (I don't know how it arranges to have appropriate alignment, but it evidently does.)

freddy77: Oh... so the bug is in structure definition. Possibly the change simply avoid the warning as linker does not check external structure alignment. Anyway, I would then add a comment about the specific MacOS issue, something like

```c
#ifdef __APPLE__
	/* some MacOS header defines gss_OID_desc with a wrong byte alignment, use external
	 * library definition. */
#  define nt_principal (*(gss_OID_desc *) GSS_KRB5_NT_PRINCIPAL_NAME)
#else
	/* same as GSS_KRB5_NT_PRINCIPAL_NAME but do not require .so library */
	static gss_OID_desc nt_principal = { 10, (void*) "\x2a\x86\x48\x86\xf7\x12\x01\x02\x02\x01" };
#endif
```

src/odbc/unittests/rowset.c
```diff
@@ -12,6 +12,20 @@ test_err(int n)
 	}
 }
 
+#ifdef _MSC_VER
+/* See https://learn.microsoft.com/en-us/cpp/preprocessor/warning?view=msvc-170 */
+#pragma warning(push)
+#pragma warning(disable:4996)
+static SQLRETURN
+_CHKSetStmtOption(SQLUSMALLINT option, SQLULEN value, const char* res)
```
freddy77, line 20: Underscore capital is reserved in C99.

That's weird having macro in `common.h` and redefinition here. Why not having 
```c
SQLRETURN SQLSetStmtOption_nowarning(SQLHSTMT hstmt, SQLSMALLINT option, SQLPOINTER param);
#define CHKSetStmtOption(a,b,res) \
        CHKR2(SQLSetStmtOption_nowarning, (odbc_stmt,a,b), SQL_HANDLE_STMT, odbc_stmt, res)
```
in `common.h` and
```c
#ifdef _MSC_VER
/* See https://learn.microsoft.com/en-us/cpp/preprocessor/warning?view=msvc-170 */
#pragma warning(push)
#pragma warning(disable:4996)
#endif
SQLRETURN
SQLSetStmtOption_nowarning(SQLHSTMT hstmt, SQLSMALLINT option, SQLPOINTER param)
{
	return SQLSetStmtOption(hstmt, option, param);
}
#ifdef _MSC_VER
#pragma warning(pop)
#endif
```
in `common.c` ?

src/dblib/bcp.c
freddy77: Just checked usage of this in db-library and in multiple cases the value can be -1 (everything less than -1 should be an error... which actually it's not properly checked :worried:). Actually if the type is fixed -1 is the value usually used!
So probably that change spot an error and code should be
```c
	if ((DBINT) bindcol->column_bindlen >= 0) {
```
wondering how many hidden bugs like this there could be in the code

ucko: In that case, perhaps the field should be signed after all.  Warnings from supplying `-1` would have been helpful, but it's apparently deemed legitimate because `-1` is a legal *signed* `int`. :-/

src/odbc/unittests/rowset.c
ucko: > Underscore capital is reserved in C99.

Ah, right.  FWIW, I'd been following the lead of all the internal `_SQL*` functions, which I suppose should be renamed to respect that reservation.

At any rate, yeah, that approach looks like it should work, though perhaps it would be better to leave `CHKSetStmtOption` as is and add a `CHKSetStmtOption_nowarning` so the test could explicitly acknowledge its use of a deprecated call.

freddy77: > > Underscore capital is reserved in C99.
> 
> Ah, right. FWIW, I'd been following the lead of all the internal `_SQL*` functions, which I suppose should be renamed to respect that reservation.

Yes, some code in FreeTDS is pretty old. C99 was surely not a thing.

> At any rate, yeah, that approach looks like it should work, though perhaps it would be better to leave `CHKSetStmtOption` as is and add a `CHKSetStmtOption_nowarning` so the test could explicitly acknowledge its use of a deprecated call.

so adding
```c
SQLRETURN SQLSetStmtOption_nowarning(SQLHSTMT hstmt, SQLSMALLINT option, SQLPOINTER param);
#define CHKSetStmtOption_nowarning(a,b,res) \
        CHKR2(SQLSetStmtOption_nowarning, (odbc_stmt,a,b), SQL_HANDLE_STMT, odbc_stmt, res)
```
in `common.h`. Sounds good to me.

src/dblib/bcp.c
freddy77: Merged everything but this hunk. Looking at all libraries it looks like all uses signed, I think it would be better to change the variable to signed.

freddy77:
Fixed last issue with https://github.com/FreeTDS/freetds/commit/bbe79e5a01ccadb65f18ee291ee958bc6fab3728

PR #582, opened 2024-05-30
ct_get_data: Distinguish NULL and empty strings via *outlen.

ucko:
Negative if NULL, 0 if merely empty.


freddy77:
There's not much testing for this function (not a regression BTW), wondering what Sybase libraries does in this case... I'll try doing some tests.

ucko:
As I recall, Sybase doesn't support non-NULL empty variable-length data in the first place, so this change only ever comes into play when working with MS SQL servers.

freddy77:
But the code here is changing the `NULL` behaviour which is supported also by Sybase.

freddy77:
Quick test. Ct-library seems to do what BCP does. For `NULL` length returned is 0, for empty length 1 is returned with a single space `' '` (`TEXT`, I suppose for `IMAGE` would return a single 0 byte, not sure other types).

ucko:
Good point.  Can you suggest an alternative way to distinguish empty columns (or tails) from both `NULL` and single spaces (or `NUL` bytes), though?  `ct_bind` takes an indicator pointer, but as a one-shot is unsuitable for blobs that could be arbitrarily long.

freddy77:
The idea of single space is that is never in the database. If you insert a string with spaces they are trimmed so it's the same as empty string. At least is what Sybase does with spaces at the end. Oh... if it's something like `CHAR(xxx)` then it's padded with spaces so you don't know if these padding where there when you tried to insert them into the database. So, no, there's no way to distinguish `''` and `' '` from database but that's expected to happen.

ucko:
As noted above, MS SQL Server distinguishes NULL data, empty strings, and single-space strings; it's just Sybase (admittedly this API's designer) that lacks a distinct representation for empty strings.

freddy77:
> As noted above, MS SQL Server distinguishes NULL data, empty strings, and single-space strings; it's just Sybase (admittedly this API's designer) that lacks a distinct representation for empty strings.

Indeed. But unfortunately both dblib and ctlib has this 0 length means NULL behaviour.

ucko:
Are you open to letting callers explicitly opt in to the proposed behavior by means of a new `ct_config` property?

freddy77:
> Are you open to letting callers explicitly opt in to the proposed behavior by means of a new `ct_config` property?

yes. So I suppose original 0 -> NULL, 1 with space -> empty, modified 0 -> empty, -1 -> NULL.

freddy77:
Can I use `9303` for `CS_NOTE_EMPTY_DATA` ? The `93xx` numbers are already FreeTDS extensions.

ucko:
> Can I use `9303` for `CS_NOTE_EMPTY_DATA` ? The `93xx` numbers are already FreeTDS extensions.

Please feel free; I'd just given it a placeholder value that wouldn't conflict with any additions from your side in the interim.  Thanks!

freddy77:
Merged

PR #583, opened 2024-05-30
Regroup pointer comparisons to avoid subtraction.

ucko:
The compiler always insists on treating the difference between two pointers as potentially negative, yielding warnings when comparing it to an integer that has (or will soon have) a pointer-width unsigned type. Equivalent pointer-to-pointer comparisons yield no such warnings and are still clear.

Split from #555.


freddy77:
Merged

ucko:
Thanks!

PR #584, opened 2024-05-30
ct_cursor: Predict cursor type for CS_CURSOR_DECLARE.

ucko:
Request forward-only cursors for updates, keyset-driven cursors otherwise. Regardless, specify optimistic cursors, which check timestamps and, when they're not available, values.

Split from #555.

### DIFF
--- a/src/ctlib/ct.c
+++ b/src/ctlib/ct.c
@@ -3900,6 +3900,29 @@ ct_poll(CS_CONTEXT * ctx, CS_CONNECTION * connection, CS_INT milliseconds, CS_CO
 	return CS_FAIL;
 }
 
+static
+char* get_next_tok(char* str, char* delimiter, char **endptr)
+{
+	char* result = NULL;
+	*endptr = NULL;
+
+	if (str && delimiter) {
+		size_t str_len = strlen(str);
+		size_t pos = strspn(str, delimiter);
+
+		if (pos == 0) {
+			*endptr = strpbrk(str, delimiter);
+			return str;
+		} else if (pos != str_len) {
+			result = str + pos;
+			*endptr = strpbrk(result, delimiter);
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


src/ctlib/ct.c
```diff
@@ -3942,6 +3965,52 @@ ct_cursor(CS_COMMAND * cmd, CS_INT type, CS_CHAR * name, CS_INT namelen, CS_CHAR
...
+				eForUpdate
+			} state = eBaseline;
+			char* savept = NULL;
+			char* s = text;
+
+			char* tok = get_next_tok(s, delimiter, &savept);
+			while (tok != NULL) {
+				s = savept;
+
+				if (strcasecmp(tok, "FOR") == 0) {
```
freddy77, line 3983: `get_next_tok` is not terminating the tokens so if you have a query like `select * from table for update` `get_next_token` will return `for update`, not `for`.

src/ctlib/ct.c
```diff
@@ -3942,6 +3965,52 @@ ct_cursor(CS_COMMAND * cmd, CS_INT type, CS_CHAR * name, CS_INT namelen, CS_CHAR
 		cursor->status.close      = TDS_CURSOR_STATE_UNACTIONED;
 		cursor->status.dealloc    = TDS_CURSOR_STATE_UNACTIONED;
 
+		if (option == CS_UNUSED	 ||  (option & CS_END) != 0) {
+			/* Try to figure out type of the cursor. */
+			char delimiter[] = "\n\t,.[]() ";
+			enum {
+				eBaseline,
+				eFor,
+				eForUpdate
```
freddy77, line 3974: camel case is a bit out of style in FreeTDS. What `e` stands for? What about `tok_baseline`, `tok_for`, `tok_for_update` ?

PR #585, opened 2024-05-30
read.c (tds_get_n): Avoid potential hangs on short replies.

ucko:
Don't bother trying to receive data following a packet marked as last; rather, bail if still short.  Whenever bailing, close the socket on the way out to contain the damage from having fallen out of sync.

Split from #555.

### DIFF
--- a/src/tds/read.c
+++ b/src/tds/read.c
@@ -240,8 +240,12 @@ tds_get_n(TDSSOCKET * tds, void *dest, size_t need)
 			dest = (char *) dest + have;
 		}
 		need -= have;
-		if (TDS_UNLIKELY(tds_read_packet(tds) < 0))
+		if (TDS_UNLIKELY(tds->recv_packet->capacity < 2
+				 ||  tds->in_buf[1] != 0
+				 ||  tds_read_packet(tds) < 0)) {
+			tds_close_socket(tds); /* evidently out of sync */
 			return false;
+		}
 	}
 	if (need > 0) {
 		/* get the remainder if there is any */


src/tds/read.c
```diff
@@ -240,8 +240,12 @@ tds_get_n(TDSSOCKET * tds, void *dest, size_t need)
 			dest = (char *) dest + have;
 		}
 		need -= have;
-		if (TDS_UNLIKELY(tds_read_packet(tds) < 0))
+		if (TDS_UNLIKELY(tds->recv_packet->capacity < 2
```
freddy77, line 243: `capacity` can't be so small

src/tds/read.c
```diff
@@ -240,8 +240,12 @@ tds_get_n(TDSSOCKET * tds, void *dest, size_t need)
...
+				 ||  tds->in_buf[1] != 0
```
freddy77, line 244: Maybe you want to test only the final bit and not all the flags. It makes sense, if no more packets are expected data should not continue on next packet. Wondering if `tds_get_n` would be called for first bytes of the packet series. Not sure if this can even happen however. I don't think cancellation apply here.

src/tds/read.c
ucko: Fair point; I just wanted to preclude any possibility that the following array access would be invalid.

src/tds/read.c
ucko: Ah, yeah, testing just the low-order bit is more appropriate, though I'm not sure how much difference it makes in practice.

The situation originally motivating this change was canceling blob retrieval, which at least in the case of the OpenServer instance I was using, could yield a truncated row response giving a blob's full length but containing only part of its body, leading to loss of synchronization that would otherwise typically culminate in hangs.  Perhaps the server should have acknowledged cancellation more clearly, but it did at least set that flag correctly.

freddy77: Interesting! So you are telling the way Open Server handles cancellation is different? Maybe we can support it.

ucko: Yes, at least historically.  Maybe it's been fixed since, but a little extra caution here can't hurt.

freddy77:
Merged, with minor changes. Sorry for the delay. One thing that puzzled me is why it did not fail when we get to a new packet; probably the packet always start with a byte read by `tds_get_byte`.

PR #586, opened 2024-05-30
tds_check_column_extra: Skip size check for BLOBs.

ucko:
Let column_size exceed column_cur_size for BLOBs, whose reported maximum may have been merely nominal, and doesn't really matter because their storage is generally allocated separately, based on actual result sizes.

Split from #555.

### DIFF
--- a/src/tds/tds_checks.c
+++ b/src/tds/tds_checks.c
@@ -235,7 +235,15 @@ tds_check_column_extra(const TDSCOLUMN * column)
 	assert(varint_ok);
 
 	assert(!is_numeric_type(column->column_type));
-	assert(column->column_cur_size <= column->column_size);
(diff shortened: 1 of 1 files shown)


freddy77:
I suppose server should not report a size lesser than the returned data (that is should consider TEXTSIZE setting). But yes, currently code won't have any issue with this and maybe is not respected for data sent to the server. Some tests here are maybe too strict. On the other side these tests should not be compiled in production and so they should not trigger and I don't see any failure during testing. Did you manage to reproduce?

ucko:
I didn't directly encounter a situation where this change helped, but AFAICT it did in fact originally compensate for a server-side bug (probably in an OpenServer instance, quite possibly in instance-specific code) that has since been fixed.  Withdrawing accordingly.

PR #587, opened 2024-05-30
tds_select: Treat EAGAIN and EINPROGRESS like EINTR.

ucko:
Call the registered interrupt handler (if any) for all three.

Split from #555.


ucko:
Thanks!

PR #588, opened 2024-05-30
odbc stats.c: Stick fully with the designated database.

ucko:
Continue to run against it when explicitly specifying a catalog (in ODBC parlance), rather than attempting to create(!) (if absent) and use freetds_test.

Split from #555.

### DIFF
--- a/src/odbc/unittests/stats.c
+++ b/src/odbc/unittests/stats.c
@@ -89,6 +89,13 @@ str(char *buf, int n)
 	return buf;
 }
 
+static void
+set_dbname(const char *dbname)
+{
+	CHKSetConnectAttr(SQL_ATTR_CURRENT_CATALOG, (SQLPOINTER) T(dbname),
+			  strlen(dbname)*sizeof(SQLTCHAR), "SI");
+}
+
 int
 main(void)
 {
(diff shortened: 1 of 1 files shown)


freddy77:
I suppose the reason for this change is to avoid the need to create a database with the required permissions associated. Or is there any other reasons? That should be stated in the commit message. Also worth saying that using default database is not limiting the effectiveness of the test (or at least it seems so looking at the code).

ucko:
Done:

> The precise database name has no bearing on this test's effectiveness.
> However, the test cases that explicitly specified a catalog (in ODBC
> parlance) to TestTable historically insisted on using freetds_test,
> and this test's only provision for lacking such a database was
> attempting to create it itself despite not necessarily having
> sufficient privileges.

On a somewhat related note (which of course deserves its own PR), it would be helpful if tests used temporary database objects to the extent possible; using nominally permanent ones is mostly no problem but can result in race conditions when running multiple tests at the same time.

freddy77:
Unfortunately there's a reason why using a different database. Queries we must send internally are different if the database is different from the current. To fix the permission issue I suppose we can assume the presence of both `master` and `tempdb` and use "the other" (based on current database) instead of `freetds_test`. About temporary tables if I remember correctly there's a reason too, related to the way database internally store temporary table names.

ucko:
Good point, though `master` may be too restricted; I've tweaked my commit to use it only as the nominal default when explicitly testing against the specified database:
> Bypass potential optimizations by formally changing to master before
> explicitly supplying the specified database name, with the help of a
> set_dbname copied from connect2.c.

freddy77:
Wonderful! Testing, did some changes (appveyor, userguide).

freddy77:
Merged, see https://github.com/FreeTDS/freetds/commit/37e68560c02d03e30dbdd87538f5dcb0debcb565

ucko:
No problem.  Thanks for pointing me in the right general direction and adding final touches!

PR #589, opened 2024-05-30
tds_fix_column_size: Allow full size for output columns.

ucko:
If an output column of type VARCHAR(n) or the like has no current size, send the type's maximum size; otherwise, the server will fill in only the first character/byte.

Split from #555.


freddy77:
This looks like a workaround for a problem in upper layer libraries. This function is supposed to make sure we don't screw with protocol level. I know ODBC is supposed to fill what it wants to be the size of the returned data, not leave as 0. I suppose that each library should detect this wrong request and report to the user. Need some more investigation.

ucko:
Ah, yeah, looks like the CTLIB API (my primary concern) works fine without this change these days, and already didn't need it in 1.0; I can't readily speak for the other two.  Withdrawing accordingly.

PR #595, opened 2024-07-03
Match Sybase dblib's handling of overlong or malformatted BCP input.

ucko:
Adjust existing failure modes' boundaries and introduce the possibility of proceeding with truncated input but reporting a message (20074, aka SYBEBCOR) as a middle ground between silent truncation and noisy failure.  To that end, pass _bcp_convert_in the destination's length.


freddy77:
This change is weird. The conditions have a final `&& 0` which will make this change a no-op. Are you still working on this?

PR #596, opened 2024-07-03
_cs_convert: Cap binary/image copying to the destination length.

ucko:
Excessively long input already yielded an error, but would previously overrun the destination buffer along the way.


freddy77:
Merged, added some tests.

ucko:
Thanks!

PR #597, opened 2024-07-03
tds_convert_to_binary: Don't leak allocations on syntax errors.

ucko:
If tds_char2hex reported a syntax error and we'd (as usual) allocated a buffer, free it on the way out so that callers won't have to account for this corner case themselves, taking care to avoid leaving a dangling pointer in the conversion-result union.


freddy77:
Changed a bit, added a test. Currently testing. Wondering how this was not detected before! Yes, the rule is that you should free result only on successful conversions. Maybe the second commit I need to backport to stable in a day! Thanks

freddy77:
Merged and backported

ucko:
Thanks!

PR #603, opened 2024-07-22
tds_lookup_host: Fall back on unsupported address families.

ucko:
If getaddrinfo with AI_ADDRCONFIG enabled fails with an error pointing to that flag (EAI_FAMILY or, if defined, EAI_ADDRFAMILY), try again without it rather than immediately bailing.  The result will be useful only in corner cases (IPv6 loopback addresses on otherwise IPv4-only hosts or vice versa), but the portconf unit test otherwise fails on IPv4-only or IPv6-only hosts supporting AI_ADDRCONFIG.


freddy77:
Merged

PR #609, opened 2024-08-26
ct: fixed logic operator precedence

ProjectMutilation:
Possible wrong assignment due to logic ops precedence. Changed by analogy to src/ctlib/blk.c

Found by Postgres Professional with ISP RAS Svace


freddy77:
Split the line and merged

PR #610, opened 2024-08-27
pool: avoided potential NULL dereference

ProjectMutilation:
The result of tds_alloc_login() is usually checked against NULL before use.

Found by Postgres Professional with ISP RAS Svace

### DIFF
--- a/src/pool/user.c
+++ b/src/pool/user.c
@@ -307,6 +307,10 @@ pool_user_login(TDS_POOL * pool, TDS_POOL_USER * puser)
 	}
 
 	puser->login = login = tds_alloc_login(true);
+	if (!login) {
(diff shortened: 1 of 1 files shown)


src/pool/user.c
```diff
@@ -307,6 +307,10 @@ pool_user_login(TDS_POOL * pool, TDS_POOL_USER * puser)
 	}
 
 	puser->login = login = tds_alloc_login(true);
+	if (!login) {
+		fprintf(stderr, "tds_alloc_login() failed.\n");
```
freddy77, line 311: I would replace `fprintf` with a log. Can I change it?

ProjectMutilation: yep

freddy77:
Changed and merged

PR #611, opened 2024-08-27
defncopy: avoid potential null dereference

ProjectMutilation:
The tds_getpassarg() function may return a NULL value, which must be carefully checked before it is used.

Found by Postgres Professional with ISPRAS Svace
Signed-off-by: Maksim Korotkov <m.korotkov@postgrespro.ru>

### DIFF
--- a/src/apps/defncopy.c
+++ b/src/apps/defncopy.c
@@ -980,6 +980,10 @@ get_login(int argc, char *argv[], OPTIONS *options)
 			break;
 		case 'P':
 			password = tds_getpassarg(optarg);
+			if (!password) {
+				fprintf(stderr, "Error getting password\n");
+				exit(1);
+			}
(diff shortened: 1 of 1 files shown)


src/apps/defncopy.c
```diff
@@ -980,6 +980,10 @@ get_login(int argc, char *argv[], OPTIONS *options)
 			break;
 		case 'P':
 			password = tds_getpassarg(optarg);
+			if (!password) {
+				fprintf(stderr, "tds_getpassarg() failed\n");
```
freddy77, line 984: I would use a more user friendly error like `"Error getting password\n"`. Can I change it ?

ProjectMutilation: fixed

freddy77 (approved):

PR #615, opened 2024-10-01
avoid version for odbc driver

fundawang:
Currently, freetds produces `libtdsodbc.so.0.0`, creates `libtdsodbc.so.0` and `libtdsodbc.so` symlinks toward it when building with autotools. But in fact, those files are not required, and this library does not need to be versioned.

So, it might be better removing version from `libtdsodbc.so`, the same as what was did when building with cmake.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -666,6 +666,11 @@ AM_CONDITIONAL(INCODBC, test "$build_odbc" = "yes")
 if test "$build_odbc" = "yes"
 then
 	case $host_os in
+	linux*)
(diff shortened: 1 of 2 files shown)


freddy77:
It looks good. Just checked also CMake on Linux do not produce such versioned files.

freddy77:
@fundawang can you add `Signed-off-by` comment? Or authorize me to do so?

fundawang:
> @fundawang can you add `Signed-off-by` comment? Or authorize me to do so?

just do it :)

freddy77:
Merged

PR #617, opened 2024-10-28
Use "TDS server" instead of "Adaptive Server" in error messages

mkauf:
FreeTDS works with both Adaptive Server and Microsoft SQL Server, so a generic term should be used that works for both servers.


freddy77:
It sounds reasonable. Launched on AppVeyor.

mkauf:
Thank you for merging!

PR #618, opened 2024-12-04
Add function to supply bulk hints to ctlib/blk feature

jboia:
There is an ability in freebcp (dblib) to specify bulk hints to the server.  This does not currently exist in ctlib/blk_*.  This patch provides a very minimal function to set the hints string in the bulk descriptor.  There is no validation.  User code should look like this:

#ifdef BLK_HINTS
    const char *hint = "CHECK_CONSTRAINTS,FIRE_TRIGGERS";
    blk_hints( blkdesc, hint );
#endif

Since it is passing a pointer, the associated string must not go out of scope until after the blk_done call is made.

Some whitespace alignment and debug cleanup are also provided for readability.

### DIFF
--- a/include/bkpublic.h
+++ b/include/bkpublic.h
@@ -36,6 +36,7 @@ extern "C"
 
 /* buld properties start with 1 i guess */
 #define BLK_IDENTITY 1
+#define BLK_HINTS    1
 
 CS_RETCODE blk_alloc(CS_CONNECTION * connection, CS_INT version, CS_BLKDESC ** blk_pointer);
 CS_RETCODE blk_bind(CS_BLKDESC * blkdesc, CS_INT colnum, CS_DATAFMT * datafmt, CS_VOID * buffer, CS_INT * datalen,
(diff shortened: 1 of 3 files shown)


freddy77:
The reason why hints are not there is mainly that Sybase/SAP does not have this feature.
Having `hint` as `const char *` is indeed a bit of a mistake, ODBC does some nasty thing to workaround this.
I tried to change it... it should work but I realized there's no testing of this feature.
I think that instead of adding a new function the proper way would be to add a property for `blk_props`. I know, that's surely a bigger change. Some example of code could be find in `ct_con_props` (`src/ctlib/ct.c`).

Could you add `Signed-off-by` line?

I think `RVSS-282` is an internal code that does not much sense here.

freddy77:
Main functionality implemented in https://github.com/FreeTDS/freetds/commit/88e0f2a2cdb925a9888bfca06cbc5e31c1cd0787.
I would pick up some changes from your commit if you don't mind. Not the ones changing formatting, code should be coherent, no assignment alignment for instance. But a couple of log changes (surely the one with the missing condition) are nice.
I would change `convert failed for srctype %d -> desttype %d` to `conversion from srctype %d to desttype %d failed`.
I won't work on this for a while, you are welcome to provide PRs.

jboia:
Thanks for adapting the suggested changes.  Feel free to improve the log messages at your discretion.  I will close this pull request.

freddy77:
Usually I don't backport new feature in stable branch but I'm not planning a release so I backported. Available in 1.4.24.

PR #619, opened 2024-12-05
Use github actions

termim:
Add github actions to compile and test (on Linux) against Microsoft SQL server.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -4,12 +4,16 @@ on:
     push:
         branches:
             - '**'
+        paths-ignore:
+            - .github/workflows/macos.yml
         tags:
             - 'v[0-9]+.[0-9]+.[0-9]+'
             - 'v[0-9]+.[0-9]+.[0-9]+rc[0-9]+'
     pull_request:
         branches:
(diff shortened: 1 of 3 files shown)


freddy77:
It seems fine. Could you add `Signed-off-by` ?

Why did you exclude `ctlib` tests ?

I noted tests are run only for Linux. Is it not possible to have 2 containers using multiple OSes ?

termim:
Added `Signed-off-by`.

I've commented out `ctlib` tests because if multiple runs are run in parallel (protocol versions * MSSQL version) `cancel.c` test hangs for 3 - 5 hours for some reason. When action is running with just one runner, then it runs fine.
Also I did not include ODBC tests as I could not properly configure it.

Unfortunately Windows and MacOS runners on Github do not support docker, so the only way to run tests on these platforms is to use some external source I guess.

freddy77:
Merged

freddy77:
The `cancel` behaviour looks weird, looking at the test it uses temporary tables, they should not conflict.

freddy77:
Fixed and enabled both ctlib and ODBC.

PR #620, opened 2024-12-11
Add release to Linux action

termim:
In Linux action create Github release when a tag starting with `v` is pushed.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -151,6 +151,24 @@ jobs:
         - name: Archive archives
           uses: actions/upload-artifact@v4
           with:
-            name: dist
+            name: dist-${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


termim:
It is possible to rerun failed actions with `Re-run` button if the failure is intermittent.

freddy77:
To be honest I would even remove `dist` job. There are packages at the official website, see www.freetds.org.

termim:
OK, no problem. You can close the PR then.

freddy77:
There are some difference why I prefer the ones from the website
- better controlled environment;
- user guide and source reference are built and packaged;
- more easy to understand which version is more recent having an incremental number.

freddy77:
As discussed removed the distribution from Github actions. Closing this PR

PR #621, opened 2024-12-12
Improve Scorecard score

minchai23:
freetds now has a score of 3.7
image
https://scorecard.dev/viewer/?uri=github.com/FreeTDS/freetds

This PR improves the scores for check items'Token-Permissions' and'Pinned-Dependencies'.

Background:
The company has a minimum rating requirement for open-source software in use; software must have a score higher than 4 to be introduced and utilized.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,6 +19,9 @@ on:
 
     workflow_dispatch:
 
+# Declare default permissions as read only.
+permissions: read-all
+
 jobs:
 
     build:
@@ -57,7 +60,7 @@ jobs:
 
         steps:
 
-        - uses: actions/checkout@v4
+        - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
         - name: Install dependencies
           run: |
@@ -112,7 +115,7 @@ jobs:
 
         - name: Archive test results on failure
           if: failure()
-          uses: actions/upload-artifact@v4
+          uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
           with:
             name: test-logs
             path: |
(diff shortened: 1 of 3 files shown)


freddy77:
Wondering why all these new "automated" tools requires always such a maintenance.

Why pinning by hash is better than pinning by version? What if a security issue is fixed? Do we need to keep security bugs?

Nothing against reducing token permissions, it makes sense.

About format format of the commits:
- please add `Signed-off-by` line;
- split by feature, not by file;
- "why?", this is a question that should be answered by the commit message (unless trivial);
- the commits should be self dependent, if I need look at the PR to understand them it means they are not self dependent

minchai23:
> Why pinning by hash is better than pinning by version? What if a security issue is fixed? Do we need to keep security bugs?
> 

Answers from ai:
```
Pinning by hash is better than pinning by version in several ways, especially when it comes to securing the software supply chain. Here are some advantages:
1. Exact Locking: Pinning by hash ensures that you are always using the exact same file, which is immutable. Even if the code in the original repository is tampered with, as long as you reference the hash, you will not be affected.
2. Preventing Unintended Updates: Version numbers can change for various reasons, such as developers releasing fixes or new features under the same version number. This can inadvertently introduce incompatible changes to your project. Hash-based pinning prevents this from happening.
3. Enhanced Security: If a dependency is found to have a security vulnerability, attackers might try to exploit it. By pinning to a hash, you ensure you're only using a known-safe version until you are ready and able to safely migrate to a patched version.
However, if a security issue is fixed, continuing to use an old hash means you are retaining a component with a known security bug. Therefore, when a security fix is available for a dependency, you should update to the new, patched version and update the hash accordingly. This way, your project benefits from the security of hash-based pinning while also applying critical security updates in a timely manner.
```

> About format format of the commits:
> 
> * please add `Signed-off-by` line;
> * split by feature, not by file;
> * "why?", this is a question that should be answered by the commit message (unless trivial);
> * the commits should be self dependent, if I need look at the PR to understand them it means they are not self dependent

I will close this PR (Pull Request) and push a new PR according to the aforementioned requirements.

PR #622, opened 2024-12-18
Update workflows to improve scorecard scores

minchai23:
1. Set default workflow permissions to read-only
2. Workflow dependencies pinned using hash values


freddy77:
Merged. Just reformatted commit messages

PR #623, opened 2024-12-19
ct_con_props: Support getting CS_ENDPOINT (socket file descriptor).

ucko:
Use Sybase/SAP ctlib's name and numeric value for the property.


freddy77:
Merged. Maybe it would be nice if trying setting it you won't get not supported property, but not a big issue

PR #624, opened 2024-12-19
ct_con_props: Support getting CS_PRODUCT_NAME.

ucko:
Sybase/SAP ctlib has no such property, so use the next available 93xx numeric value.


freddy77:
Merged manually. Fixed formatting and added check for TDS state (can be even NULL).

PR #625, opened 2024-12-19
Introduce CS_INTERRUPT_CB and corresponding return values: CS_INT_*.

ucko:
Define CS_INTERRUPT_CB's numeric value alongside other CS_*_CBs'; start a new 94xx range to preclude accidental duplication. (Sybase/SAP ctlib has no such property.)  For CS_INT_*, copy TDS_INT_* and confirm their equality at compile time as with CS_NULLTERM et al.


freddy77:
The pointer to pointer to function looks odd to me. In `ct_callback` `func`, passed as `CS_VOID *` is simply converted to function pointer. About `ct_callback`: won't be sensible to extend `ct_callback` instead of `ct_config` ?

ucko:
Good points; I've reworked it as an extension to `ct_callback`, which I'd somehow overlooked.

ucko:
I've force-pushed a new version of this patch addressing your feedback (apart from the very longstanding use of underscore prefixes in  `_*_cb` field names).

freddy77:
Merged. Indented test source again.

PR #628, opened 2025-01-22
Fix handling of SQL_C_STINYINT

### LINKED ISSUE #627: Negative `SQL_C_STINYINT` values don't seem to be handled correctly
Using `SQLBindParameter(SQL_C_STINYINT, SQL_SMALLINT)` for inserting into a `SMALLINT` column doesn't seem to work correctly: if the parameter value is negative, it's stored in the database as a positive value as if it were cast from `signed char` to `unsigned char`, e.g. -5 is saved as 251.

This doesn't happen with `SQL_C_SSHORT, SQL_SMALLINT` nor when using Microsoft ODBC driver.

vadz:
Map it to SYBSINT1 so that it is stored as signed, and not unsigned as before, in the database.

Closes #627.


freddy77:
Can you add a `Signed-off-by` line?

vadz:
Sure, force-pushed now, but it's not mine patch anyhow — I've just made the PR to run the CI jobs, but I thought you'd just cherry pick this commit with your authorship because you wrote it and not I.

freddy77:
Okay, I'll write one then, I have updated a test too and tested with MS drivers, just in case.

PR #629, opened 2025-01-22
Let array test pass with Microsoft ODBC driver

vadz:
I'd like the test to pass with this driver too to be able to compare behaviour between FreeTDS and it easier, so I hope these changes could be applied (if not, I'll just keep them in a local branch but this will mean more juggling with branches for me).


freddy77:
In theory the behaviour should be the same... but there are indeed differences.
I'll merge.
Yes, it looks like MS driver has multiple bugs like this :cry:

PR #632, opened 2025-01-24
Fix return value of bulk insert without status array

### LINKED ISSUE #626: Behaviour differs from Microsoft ODBC driver for bulk insert in case of an integrity violation error
Suppose you initialize an empty database in the following way:

```sql
create table parent(id integer primary key);

create table child(id integer primary key, parent integer,
    foreign key(parent) references parent(id)
);

insert into parent(id) values(1);
insert into child(id, parent) values(100, 1);
```

Then executing

```sql
insert into child(id, parent) values(:child, :parent)
```

with `child` bound to an array containing `{200, 300, 400}` and `parent` bound to an array containing `{1, 2, 1}` fails, as expected, with Microsoft ODBC driver due to the foreign key constraint violation for the second element, i.e. `SQLExecute()` returns an error and `SQLGetDiagRec()` returns the expected error:

```
[Microsoft][ODBC SQL Server Driver][SQL Server]The INSERT statement conflicted with the FOREIGN KEY constraint "FK__child__parent__24DD5622". The conflict occurred in database "soci_test", table "dbo.parent", column 'id'.
```

However, when using FreeTDS (tested with both 1.3.17 from the current Debian Stable and current master which is `R1_4RC1-375-gd393b611`), the same statement does _not_ fail, `SQLExecute()` returns `SQL_SUCCESS_WITH_INFO` and the database _is_ modified, with the second element somehow being inserted with the parent of `1`, which is very unexpected. There is still the associated error message:

```
[FreeTDS][SQL Server]The INSERT statement conflicted with the FOREIGN KEY constraint "FK__child__parent__2018A105". The conflict occurred in database "soci_test", table "dbo.parent", column 'id'.
```

but it is not really used because the statement is executed successfully and, anyhow, the database has been modified, so reporting this as an error would be misleading.

Is there some way to make FreeTDS behave in the same way as Microsoft ODBC driver in this case? I.e. to make it fail with the foreign key constraint violation and not modify the database?

vadz:
Extend the unit test to check for this case.

Fixes #626.

### DIFF
--- a/src/odbc/unittests/array.c
+++ b/src/odbc/unittests/array.c
@@ -9,8 +9,12 @@ static int failure = 0;
 
 #define XMALLOC_N(t, n) (t*) ODBC_GET(n*sizeof(t))
 
+enum {
+	FLAG_PREPARE = 1
+};
+
 static void
-query_test(int prepare, SQLRETURN expected, const char *expected_status)
+query_test(int flags, SQLRETURN expected, const char *expected_status)
 {
 #define DESC_LEN 51
 #define ARRAY_SIZE 10
@@ -50,7 +54,7 @@ query_test(int prepare, SQLRETURN expected, const char *expected_status)
 	}
 	multiply = 90;
 
-	if (!prepare) {
+	if (!(flags & FLAG_PREPARE)) {
 		ret = SQLExecDirect(odbc_stmt, test_query, SQL_NTS);
 	} else {
 		SQLPrepare(odbc_stmt, test_query, SQL_NTS);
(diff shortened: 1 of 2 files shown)


src/odbc/unittests/array.c
```diff
@@ -107,6 +116,10 @@ query_test(int prepare, SQLRETURN expected, const char *expected_status)
 			exit(1);
 		}
 
+		/* can't check status values if we hadn't asked for themm */
```
freddy77, line 119: typo: themm -> them

vadz: Thanks for noticing, will (force) push to fix this soon (when the CI jobs finish, just to see if there are no unforeseen problems).

PR #634, opened 2025-02-04
dblib: fixed incorrect error handling

ProjectMutilation:


src/dblib/dblib.c
```diff
@@ -725,6 +725,7 @@ dblogin(void)
 	/* set default values for loginrec */
 	if (!tds_set_library(loginrec->tds_login, "DB-Library")) {
 		dbperror(NULL, SYBEMEM, errno);
+		free(loginrec->tds_login);
```
freddy77, line 728: I think `tds_free_login` would be better here, more symmetrical.

ProjectMutilation: thanks, fixed

PR #636, opened 2025-02-06
pool: added missing deallocations

ProjectMutilation:


ProjectMutilation:
@freddy77 could you please help with the review?

freddy77:
> @freddy77 could you please help with the review?

What do you mean ? Any doubt ?

ProjectMutilation:
> > @freddy77 could you please help with the review?
> 
> What do you mean ? Any doubt ?

I'm not sure if when a tds_alloc_connection() is failed inside tds, the conn->tds_ctx is not released somewhere and `tds_free_context(context);` will not be the reason for the double free

freddy77:
> > > @freddy77 could you please help with the review?
> > 
> > 
> > What do you mean ? Any doubt ?
> 
> I'm not sure if when a tds_alloc_connection() is failed inside tds_alloc_socket(), the conn->tds_ctx is not released somewhere and `tds_free_context(context);` will not be the reason for the double free

Fine as it is.

PR #644, opened 2025-03-27
More robust filename comparison on log_elision.c

craigberry:
__FILE__ on VMS includes the absolute path and version number, like so:

DKA0:[CRAIG.freetds-1^.5rc2.src.tds.unittests]log_elision.c;1

which was not matching the sscanf pattern in the test. So just concatenate __FILE__ (after expansion) onto the parts that are stable.

### DIFF
--- a/src/tds/unittests/log_elision.c
+++ b/src/tds/unittests/log_elision.c
@@ -35,6 +35,8 @@ enum {
 	THREADS = 3,
 };
 
+#define CONCAT(a, b) a##b
+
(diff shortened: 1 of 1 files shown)


freddy77:
Testing a different approach, see https://github.com/FreeTDS/freetds/commit/494d32aa6a8f20cecc5f967a09554eaea1981b59. `num_line` was not used and somebody could remove `TDS_DBGFLAG_SOURCE` setting.

craigberry:
cb8eb2a7a44 looks like a better way to do it, so I'll close this PR.

Issue #651, opened 2025-04-22
freebcp to support ASE datarows-locked tables


mmcnabb-vms:
Sybase / SAP ASE has a table lock mode called DATAROWS . If a table is defined in this way, then bulk loading requires the packed row data to be formatted differently. (Bulk unloading is unchanged).

We need to send a probe to the server first to detect if the table is datarows-locked:  

    select sysstat2 from database.owner.sysobjects where type='U' and name='tablename'

This returns an integer where the flag 0x8000 is set if the table uses DATAROWS locking.  These flags are documented in [System Tables](https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc36274.1550/html/tables/X14933.htm).

The packed format has an extra 4 bytes at the start (after the length header but before the variable column count); and the variable column offset table is in a newer format.

freddy77:
Fixed

PR #652, opened 2025-04-22
Issue 651 - freebcp support datarows-locking in ASE

mmcnabb-vms:
Sybase/SAP ASE has a table option called DATAROWS locking (the default is ALLPAGES). If this is set, then freebcp needs to format the bulk row data differently. 

This means that on freebcp startup, we must probe the table to see what its locktype is by checking `sysstat2` for the table -- I got this by sniffing the connection for ASE 16.0's bcp.exe utility.

For this change I have limited the probe to occur for ASE versions that support it. This would cause a small runtime penalty for freebcp on ASE systems that aren't using datarows locking; it could be possible to add a mechanism whereby the freebcp user can specify at runtime whether or not they want the TDS_REQ_DOL_BULK capability to be flagged.

The ASE 16.0 bcp utility always carries out this probe even if the server doesn't have any datarows-locked tables active.

I have tested this by using the `c_blk_in` unittest but adding `LOCK DATAROWS` to the end of the `CREATE TABLE` for the first table.   Is it possible to add a dedicated unit test that only runs if we're talking to a recent ASE version that has this feature - please let me know how to go about this if so.


freddy77:
Can you add `Signed-off-by` tag in the commit messages?

mmcnabb-vms:
@freddy77 I haven't worked with signoff before -- do you want me to rewrite my whole PR branch with --signoff on every commit?   Are there any other project policies I should check before doing this (e.g. code style)

freddy77:
Don't worry about style, I can adjust along the way. But the `Signed-off-by` is kind of required, it stated that you accept the license and give the code to the project.

mmcnabb-vms:
@freddy77 OK, I have force pushed with signoff. I used: git rebase --signoff HEAD~7 .  Will do the same for the SSL issue (511) once we decide what to do about the certificate check.

freddy77:
Merged. Not sure about the tests. Indeed they would help. Other bulk tests check if the creation fails and skip the test on error.

mmcnabb-vms:
@freddy77 OK, you will close #651 and #652 then ? And I will make a new PR if adding a test case.

freddy77:
Added a test, no regressions

Issue #511, opened 2023-10-30
How to connect to a Sybase SSL port if the Sybase server uses TDS version 5.0?


sagarmavuri:
How to connect to a Sybase SSL port using `FreeTDS`? I am using `pyodbc` and while I am able to connect to a non-SSL port just fine, but SSL connections aren't working. 

Note that the Sybase server uses TDS version 5.0.

sagarmavuri:
Also, I have used FreeTDS 1.1.6 and 1.4 as well. Nothing works, when I look at the C++ files, the SSL checks are only enabled for TDS version 7 or greater. 
So when I use TDS version 7 or more, the Sybase server is rejecting the connections. How do I make it this to work?

freddy77:
It sounds like something not supported by FreeTDS. Can you do a network trace of a successful connection (that is using Sybase libraries) to a SSL port? If you are concerned about security issues use a test server and/or test account/password and/or send privately to me.

mmcnabb-vms:
It works to connect through Stunnel .  Otherwise, code changes are needed.

FreeTDS Branch 1.5 introduced a mode `TDS_ENCRYPTION_STRICT`, which is exactly what is needed for Sybase-SSL. However the code mistakenly assumed nobody would use this mode outside of TDS 8.0+ , so it's not possible to enable it for a TDS 5.0 server.

I found I was able to get it working by forcing the server to TDS 5.0 in `freetds.conf` (instead of autodetect), and the following changes:

```
------------------------------ src/tds/config.c -------------------------------
index f83d92cd..c77ac662 100644
@@ -1461,7 +1461,9 @@ tds_get_compiletime_settings(void)
 TDSRET
 tds8_adjust_login(TDSLOGIN *login)
 {
-	if (!IS_TDS80_PLUS(login) && login->encryption_level != TDS_ENCRYPTION_STRICT)
+	// (MM) Bugged - TDS_ENCRYPTION_STRICT should not force version 8.0 if a version was already set
+	//if (!IS_TDS80_PLUS(login) && login->encryption_level != TDS_ENCRYPTION_STRICT)
+	if (!IS_TDS80_PLUS(login))
 		return TDS_SUCCESS;
 
 	login->tds_version = 0x800;

------------------------------- src/tds/login.c -------------------------------
index 57786ce3..4e4c36e6 100644
@@ -687,6 +687,11 @@ reroute:
 		db_selected = true;
 	} else {
 		tds->out_flag = TDS_LOGIN;
[...]

PR #653, opened 2025-04-22
Support Sybase/ASE SSL

mmcnabb-vms:
See [issue 511](https://github.com/FreeTDS/freetds/issues/511)

Sybase SSL behaves like cleartext Sybase behind a Stunnel.

Added in that the TDS 5 login will honour TDS_ENCRYPTION_STRICT; but also had to modify `tds8_adjust_login()` to not perform adjustments if it's TDS 5.

Note that the CA certificate step is NOT necessary for ASE (and in fact breaks the handshake if I add that in, but that could also be due to a configuration error in my setup).

### DIFF
--- a/src/tds/config.c
+++ b/src/tds/config.c
@@ -1465,16 +1465,21 @@ tds_get_compiletime_settings(void)
 TDSRET
 tds8_adjust_login(TDSLOGIN *login)
 {
-	if (!IS_TDS80_PLUS(login) && login->encryption_level != TDS_ENCRYPTION_STRICT)
-		return TDS_SUCCESS;
+	// TDS 8.0 requires TDS_ENCRYPTION_STRICT (entire-connection encryption)
+	if (IS_TDS80_PLUS(login))
+		login->encryption_level = TDS_ENCRYPTION_STRICT;
 
-	login->tds_version = 0x800;
-	login->encryption_level = TDS_ENCRYPTION_STRICT;
+	if (login->encryption_level == TDS_ENCRYPTION_STRICT)
+	{
+		// TDS 5.0 (Sybase/SAP ASE) it is optional; but TDS 7.x does not allow it.
+		// So, try 8.0 unless they specifically requested 5.0
+		if (!IS_TDS50(login) && !IS_TDS80_PLUS(login))
+			login->tds_version = 0x800;
 
-	/* we must have certificates */
-	if (tds_dstr_isempty(&login->cafile)) {
-		if (!tds_dstr_copy(&login->cafile, "system"))
-			return -TDSEMEM;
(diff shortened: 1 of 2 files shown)


src/tds/config.c
```diff
@@ -1465,16 +1465,18 @@ tds_get_compiletime_settings(void)
...
-		if (!tds_dstr_copy(&login->cafile, "system"))
-			return -TDSEMEM;
+	if (login->encryption_level == TDS_ENCRYPTION_STRICT)
+	{
+		// If a version is not set yet, then try 8.0  (TDS 7.x doesn't support this).
+		// SAP/Sybase ASE also supports this mode however.
+		if (!IS_TDS50(login) && !IS_TDS80_PLUS(login))
+			login->tds_version = 0x800;
+
+		// 8.0 Strict Encryption requires certificates
```
freddy77, line 1475: Are you telling that Sybase does not mandate to specify a certificate allowing man-in-the-middle?

mmcnabb-vms: I have checked, and their client (isql) does validate the server certificate's signature against a local file of trusted CA public keys. If you don't give a server CA file then the connection fails . 

My test setup is that I have created a self-signed server cert; and have NOT imported that into the Windows cert store; I am just using the local `trusted.txt` file as recommended by ASE. I don't know if isql also checks the Windows cert store.

I have found that FreeTDS tsql also connects successfully if I specify in freetds.conf:

    ca file = C:\path\to\trusted.txt
    check certificate hostname = false

(since my test server isn't on a hostname matching the self-signed cert).

---

The code in `tds_ssl_init()` will only enable `SSL_VERIFY_PEER` if `ca file` is non-empty -- it will happily carry on and connect if no `ca file` was set.  So the user of FreeTDS can choose whether to verify server certificate (i.e. whether to try and prevent MITM or not).  

Are you suggesting we should do the `cafile` check from `tds8_adjust_login()` in all cases of Strict? If so, then I would suggest we hoist that check out of `tds8_adjust_login()`  (since it's not specific to TDS8), perhaps it can go in `tds_connect_and_login`? I can revise my PR accordingly.

mmcnabb-vms: @freddy77 On review, my PR has an unintended change -- your code forces STRICT if TDS 8.0 is set in config . I will revise my PR (& sign this time)

mmcnabb-vms: Pushed and signed now ( & rebased to master)

mmcnabb-vms: @freddy77 just bumping this one... revised PR look acceptable? (commit 971a42). Or make any change you like, so long as we can still enable TDS_ENCRYPTION_STRICT for ASE in some way

freddy77: It was fine, just busy. Merged (changed some lines for style).

PR #660, opened 2025-06-23
Better test for vasprintf on VMS

craigberry:
Recentish CRTLs do have it, but the test I added 23 years ago was not correctly detecting it.


freddy77:
Done, backported too.
Thanks

Issue #657, opened 2025-06-12
FreeTDS is Ignoring GNUTLS Configuration(s)


hking-forescout:
On a RHEL9 server I have FreeTDS 1.3.3 installed and configured with GNUTLS. I am encountering the issues listed below related to FreeTDS gnutls configuration.

1) I have tried updating the supported ciphers and priority string being used by my application by updating the system config file. I confirmed in the dump file that it is referencing that config file, but it said it was not updated after my changes even after rebooting the server. I added in a custom config file and used the `GNUTLS_SYSTEM_PRIORITY_FILE `environment variable at the start of my application runtime to reference my custom gnutls config file. This was recognized by the dump file, but any later changes to this file are also not recognized and the log says it is unchanged.

2) FreeTDS is having difficulty with the gnutls priority string. I tried enforcing different priorities with the `GNUTLS_PRIORITY `environment variable, but this is not being recognized whether I provide a string listed in the [ priorities ] section or provide a priorities string directly. The dump file always says it is loading SYSTEM priorities. I updated the [ priorities ] section `SYSTEM` string in every gnutls.config file on the system (the original and custom ones) but this is not recognized either. I see in tls.c starting at line 513 (566 in master as of 6/12/2025) we are using the `gnutls_set_default_priority(session)` call then immediately overwriting them depending on tls version. Is there a reason for this?

3) The dump file shows that the number of protocols, ciphersuites, signatures, etc. are all aligned with what is in the configuration file. Then immediately after this is replaced and freeTDS is loading the full suite that was originally in the system. Wireshark shows the full set is being passed in the tls negotiation. Is this solely related to the defaults mentioned in point 2?

My teammate found a resource that suggested that freetds may initilize a system default version when it is installed/configured with gnutls and that may also be what is overriding all of our attempts to customize our gnutls behavior. I haven't found anything in the source code that has confirmed this so does any developer know if this is accurate and a potential contributor of these issues?

**_TL;DR:_** in tls.c freetds is overriding some gnutls preferences by default. Is there a potential work around to ensure gnutls config file settings are used instead or is customization the only option?

OS: RHEL9
FreeTDS [1.3.3](https://www.freetds.org/files/stable/freetds-1.3.3.tar.bz2)
GNUTLS: 3.7.1
TDS Version: auto (we're setting 7.4 in freetds.conf)
Source code referenced from tag zip file: https://github.com/FreeTDS/freetds/releases/tag/v1.3.3

[...]

PR #662, opened 2025-07-01
tds: Add gnutls ciphers parameter to control priority string directly

hking-forescout:
Adding in the ability to directly control the gnutls priority string via the added `gnutls ciphers` parameter within freetds.conf file. This is following the pattern of `openssl ciphers`. If this parameter is not defined within in the freetds.conf file generate the gnutls priority as previously defined.

This is related to [Issue 657](https://github.com/FreeTDS/freetds/issues/657)


freddy77:
Rebased and merged, thanks

Issue #680, opened 2025-10-28
ODBC - Re-executing a prepared SELECT after SQLCloseCursor can fail if not all rows were fetched


mmcnabb-vms:
The following sequence is [supposed to work](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlclosecursor-function?view=sql-server-ver17):

- `SQLAllocHandle` (for statement)
- `SQLPrepare`
- `SQLExecute` (a SELECT query returning multiple rows)
- `SQLFetch` (called some number of times, but not enough times to fetch all of the rows)
- `SQLCloseCursor`
- `SQLExecute`
- `SQLFetch` (until all rows fetched)

However my test results, in Linux, indicate:

- Always works with MS native ODBC driver for Linux (ver 18)
- Sometimes fails for FreeTDS, with the second SQLExecute giving SQLSTATE `HY000` "Unknown error", native error `0` (no further diagnostics).
- Always works with FreeTDS and `SQLExecDirect` (instead of `SQLPrepare` and `SQLExecute`)

In the failure case, the size of the unread rows seems to matter. I am testing with fetching rows from a table of varchar columns total width about 140, and find:
- Fetch 5 rows out of 10 always succeeds
- Fetch 10 rows out of 1000 usually fails
- Fetch 999 rows out of 1000 always succeeds
- Fetch 10 rows out of 100 fails most of the time and sometimes succeeds
- If TDSDUMP is enabled, it seems to always succeed

Test conditions:
 - current "master" head , with `--enable-odbc`  (but not enable MARS ; however the problem does also occur on MARS builds)
 - unixODBC 2.3.11 (distro packaged).
 - Server and client on different machines, although network latency is under 1ms

mmcnabb-vms:
Analysis of the above logs:

- Both cases have `util.c:83:logic error: cannot change query state from IDLE to PENDING`
- The success case includes (among others) the following lines, that the failure case does not: `ff(DONEINPROC)`, `rows_affected = 100`, `calling tds_process_param_result`.

mmcnabb-vms:
Here is a diff of failure versus success logs for a 1000-row select with 100-row fetch (failure cases are first):

```
[Omitted earlier lines where the differences are just handle values, and the encrypted handshake]

1006c1006
< token.c:558:tds_process_tokens(0x188e040, 0x7ffca06520fc, (nil), 0x0)
---
[...]

PR #681, opened 2025-11-01
ODBC: Re-prepare a prepared query if it never returned a handle

mmcnabb-vms:
As discussed on https://github.com/FreeTDS/freetds/issues/680


freddy77:
Merged and backported.

Issue #679, opened 2025-10-27
Add support for EPA


paraddise:
**Is your feature request related to a problem? Please describe.**
Microsoft added `Extended Protection for Authentication` to MSSQL, so client must provide channel binding in NTLM or Kerberos authentication request.

**Describe the solution you'd like**
Library always send channel binding when authenticating via NTLM and Kerberos

**Describe alternatives you've considered**
In MSSQL we can request, accept or ignore channel binding. If one client doesn't support CBT we cannot set it to require.

**Additional context**
[Extended Protection for Authentication Overview](https://learn.microsoft.com/en-us/dotnet/framework/wcf/feature-details/extended-protection-for-authentication-overview)

[How to Easily Bypass EPA to Compromise Any Web Server that Supports Windows Integrated Authentication](https://www.crowdstrike.com/en-us/blog/bypass-epa-ntlm-attacks-wia/)

[Ldap3: commit for TLS channel binding support](https://github.com/ly4k/ldap3/commit/87f5760e5a68c2f91eac8ba375f4ea3928e2b9e0#diff-c782b790cfa0a948362bf47d72df8ddd6daac12e5757afd9d371d89385b27ef6R1383)

[A journey implementing Channel Binding on MSSQLClient.py](https://sensepost.com/blog/2025/a-journey-implementing-channel-binding-on-mssqlclient.py/)

[Impacket PR: MSSQL Implements Channel Binding Token](https://github.com/fortra/impacket/pull/1986)

MsvAvChannelBindings must be in `AV_PAIR` that can be in `CHALLENGE_MESSAGE` or in `AUTHENTICATE_MESSAGE` via  `NTLMv2_CLIENT_CHALLENGE`.
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/83f5e789-660d-4781-8491-5f8c6641f75e

[Certipy implementation of channel binding](https://github.com/ly4k/Certipy/blob/a80fe7c88d9e1a6c4cfec3732fca9ec440d5ed9d/certipy/lib/channel_binding.py#L50)

I also want to mention that many offensive tools already has support for EPA.

**Possible solution**
I see channel binding disabled when calling `gss_init_sec_context`, maybe we can enable it here?
https://github.com/FreeTDS/freetds/blob/master/src/tds/gssapi.c#L398

freddy77:
Merged the PR

PR #682, opened 2025-11-01
Add support for EPA

paraddise:
#679
Implement for NTLM:
- [x] OpenSSL
- [x] GnuTLS
- [x] tests

Implement for Kerberos:
- [x] OpenSSL
- [x] GnuTLS

Implement for SSPI
- [x] OpenSSL
- [x] GnuTLS
 
Analyze the changes needed for TDS 8.0

### DIFF
--- a/src/tds/sspi.c
+++ b/src/tds/sspi.c
@@ -40,6 +40,14 @@
 #if HAVE_SSPI
 #define SECURITY_WIN32
 
+#if defined(HAVE_OPENSSL)
+#include <openssl/ssl.h>
+#endif
+
+#if defined(HAVE_GNUTLS)
+#include <gnutls/gnutls.h>
+#endif
+
 #include <freetds/windows.h>
 #include <security.h>
 #include <sspi.h>
@@ -65,6 +73,13 @@
  */
 
 #define PRIsSTR "S"
+#define TDS_CALC_CB_SIZE(cb) \
+    ((cb) == NULL ? \
+        0 : \
+        ( sizeof(SEC_CHANNEL_BINDINGS) + \
+          (cb)->cbInitiatorLength + \
+          (cb)->cbAcceptorLength + \
+          (cb)->cbApplicationDataLength ))
 
 typedef struct tds_sspi_auth
 {
@@ -72,6 +87,7 @@ typedef struct tds_sspi_auth
 	CredHandle cred;
 	CtxtHandle cred_ctx;
 	TCHAR *sname;
+	PSEC_CHANNEL_BINDINGS cb;
 } TDSSSPIAUTH;
 
 static HMODULE secdll = NULL;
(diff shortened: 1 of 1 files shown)


paraddise:
@mmcnabb-vms Can you look at this? Will this be merged at all or am I trying in vain? 
If Yes, write please the plan what I need to add else to this PR. tests? implementation for other ssl library?

freddy77:
Currently testing changes to `src/tds/challenge.c` (from Linux).

Work:
- TDS 7.4, required encryption, GnuTLS;
- TDS 7.4, required encryption, OpenSSL (minor code change in order to compile).

Do no work:
- required encryption off. At the time code tries to get the channel binding there's no TLS objects;
- TDS 8.0, probably requires `tls-exporter` instead of `tls-unique`.

I still need to setup an AD domain to test GSSAPI.

freddy77:
> I can't get channel binding working for TLS 1.3. First, I wrote [a golang implementation](https://github.com/microsoft/go-mssqldb/pull/295/commits/8699ca8556497b495afca7908af796c97673ae98#diff-1afc8be09e03ff2c43558754d36faa5d6c26a911f1ed5cea4871d1c1a824dc0dR185) of the tls-exporter. The channel binding implementation was simple, but I didn't understand how Microsoft calculated it. Do you know which DLL or binary I could reverse to understand the secret?
> 

I suppose then your code it's not working. Looking around I would look at "token bindings" (rfc 8471) instead of tls exporter and rfc 9266. The DLL should be `schannel.dll` but please don't do it, I won't like to go to legal issues.

> I also want to point out that we need to test this on fresh MSSQL servers, as Microsoft Server 2019 does not support TLS 1.3 and when you use the encrypt=strict option, you will be using TLS 1.2 instead. So it's better to test 2019 and 2022 on Winwdows Server 2019 and 2022
> 

I'm getting TLS 1.3 on w2022, just adding some registry keys.

> And very old mssql servers have channel binding checks broken, they ignore them even if you set Enhanced Protection = Required

I don't think we can do anything about it. Unless we suggest to just force encryption at the client level.

freddy77:
> Maybe we can skip this for tds 8.0 and tls 1.3? I could not find the project that implemented it. I think we can create issue and wait until microsoft will issue some documentation how to build channel building structure with tls 1.3 with InitializeSecurityContext.

Yes, already working on it, testing is part one.
Looking at the way you did the commit I suppose I can pretty much merge them.
I can came out with commit message, or you can suggest.
I'll try to prepare and send what I'll get when I finish.

About TLS 1.3 is even possible that simply they "trust it", in the sense that with strict mode the certificate is required and all encrypted too so you can't have much man in the middle.

freddy77:
The proposed version is at https://github.com/freddy77/freetds/commits/epa/.

Changes I did:

- reduced differences removing indentation changes;
- move common code to `src/tds/tls.c` and `include/freetds/tls.h`;
- removed some allocations and leaks;
- do not mix declaration and code, some compiler does not like this;
- send channel binding on NTLM even if encryption is not set to `force` or `strict`;
- fixed minor compiler issues;
- add license and description to test;
- better check for target info from server.

You can see the full changes at https://github.com/freddy77/freetds/commits/epa0/ (before reorganizing the series).
The commit have your name, I retained the ****.qemu email. Is that correct?
Commit https://github.com/freddy77/freetds/commit/a78149e810a0a1b7945b4d4660e454677526538b has most information, more or less taken from your issue report and rephrased.

freddy77:
Merged. Thanks

PR #687, opened 2025-12-03
VMS - build system, ODBC and tests

mmcnabb-vms:
The following all apply to VMS builds:

- ODBC build now works.
- Ensure that tests actually report failure when they fail.  
- Apps (e.g. freebcp) to exit to operating system with VMS failure status. (exit status 1 is success in VMS, but there's a compiler switch to translate exit status from POSIX-style code)
- Find driver for DriverConnect ODBC tests
- All tests working, tested against SQL Server, and ASE
- Option to enable ODBC MARS and accompanying test
- Option to enable ODBC WIDE
- VMS ODBC build did not work prior to this. Have got ODBC build working, and all tests pass , tested against SQL Server, and ASE 16. 
- Fix potential problem with config.h and defining sizes of types for VMS  (different compilers use different sizes in some cases)
- Tidy up build documentation
- Add extra configure script to configure from trunk (i.e. a github checkout), the existing one only worked on release tarball.

Should be no impact on anything outside of VMS-sphere (as long as the .gitignore doesn't cause trouble!)


include/odbcss.h
```diff
@@ -266,6 +266,8 @@ typedef struct
 
 #ifdef _MSC_VER
 #define TDSODBC_INLINE __inline
+#elif(defined __VMS)
+#define TDSODBC_INLINE inline
```
freddy77: Would something like this work?
```c
#if defined(__cplusplus) || (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L)
#define TDSODBC_INLINE inline
#elif defined(_MSC_VER)
#define TDSODBC_INLINE __inline
#else
#define TDSODBC_INLINE __inline__
#endif
```

src/odbc/unittests/common.c
```diff
@@ -93,7 +94,18 @@ odbc_read_login_info(void)
 	if (!PWD && !read_login_info_base(&common_pwd, "PWD"))
 		return 1;
 
-	/* find our driver */
+	/* 
+     * Some of the ODBC tests "go in the front door", i.e. call
+     * SQLDriverConnect() which will look up an entry in ODBCINST.INI .
+     * In CONNECT and TIMEOUT3 they look up the entry FreeTDS.
```
freddy77: Capitalized names looks odd for me. Would `"connect"` suit too? I think `ODBCINST.INI` is fine.

src/odbc/unittests/connect.c
```diff
@@ -80,6 +80,9 @@ TEST_MAIN()
 	printf("odbcinst.ini must exist and contain an entry [FreeTDS].\n");
 #endif
 	/* this is expected to work with unixODBC, and Windows.
+     * NOTE: This will connect to whatever driver you have configured in ODBCINST.INI,
+     * not the actual driver we just built and are testing. However for this test CONNECT,
```
freddy77: `CONNECT` here looks redundant, we are in that specific test.

vms/config_h.vms
```diff
@@ -268,29 +268,53 @@
...
+
+/* The size of 'void *', as computed by sizeof. */
+/* May vary based on compiler setting in VMS -- Not used in codebase unless __hpux__ anyway */
+#undef SIZEOF_VOID_P
+
+/* The size of 'wchar_t', as computed by sizeof. */
+#define SIZEOF_WCHAR_T 4
+
+/* The size of '__int128', as computed by sizeof. */
+#undef SIZEOF___INT128
```
freddy77: What are the compilers used in OpenVMS? Maybe something like
```c
#if defined(__GNUC__) && defined(__SIZEOF_INT128__)
#define SIZEOF___INT128 __SIZEOF_INT128__
#endif
```
could help?

vms/generate_version.pl
```diff
@@ -0,0 +1,68 @@
+
+#!/usr/bin/perl
```
freddy77: I think the shebang line has to be the first one.
Why not `#!/usr/bin/env perl` ?

vms/config_h.vms
craigberry: There is no int128_t in stdint.h and I haven't heard of any compiler on VMS that supports it.  Probably best to leave SIZEOF___INT128 undef for now.  I think for SIZEOF_LONG_INT there would be no harm in:

```C
#define SIZEOF_LONG_INT sizeof(long int)
```

craigberry:
> I think most of this PR will need @craigberry to have a look.

It looks good and builds fine for me, but I haven't tinkered with unixODBC in probably 20 years (no surprise there has been some bitrot in ODBC support).  So I need to get unixODBC built, build FreeTDS to use it, and then run tests. I will try to get to that soon, but since the PR gets us from ODBC support on VMS definitely not working to almost certainly working, I don't know that it needs to wait for me.

vms/config_h.vms
mmcnabb-vms: There's no gcc port , the only compiler this will be built with is VSI C

vms/generate_version.pl
mmcnabb-vms: Shebangs mean nothing to the VMS command interpreter, that's no `/usr/bin` anyway. You could probably remove that line, or change it as you see fit if you wanted to re-use this script for other platforms. We invoke perl scripts via `perl filename.pl` in a command script.

vms/config_h.vms
mmcnabb-vms: > There is no int128_t in stdint.h and I haven't heard of any compiler on VMS that supports it. Probably best to leave SIZEOF___INT128 undef for now. I think for SIZEOF_LONG_INT there would be no harm in:
> 
> ```c
> #define SIZEOF_LONG_INT sizeof(long int)
> ```

These macros are for use in preprocessing conditionals (because during the compilation phase `sizeof` can be used); if defined  they have to evaluate to a number during preprocessing . Of course it would be good to minimize their necessity or ideally not use them at all . 

UnixODBC headers do unfortunately use `SIZEOF_LONG_INT` at this point in time , and in fact have a `#ifndef SIZEOF_LONG_INT` block, so that the size of types and struct layouts defined by those headers can vary based on whether you include `sql.h` before or after FreeTDS headers. So we can't set it to an incorrect value ,  I think it is the most stable option to just not define it. FreeTDS doesn't appear to need this value.

freddy77: As the comment on `configure.ac` says
```sh
# This will define SIZEOF_LONG_INT, used by some ODBC Driver Manager
```
Specifically unixODBC. We include the headers and some version seem to require this definition.
FreeTDS don't use it directly but needs correct definitions from unixODBC.

freddy77: Supposing the the value is either 4 or 8 and looking at `vmc/configure.com` would not something like this
```
$!
$! Check for sizeof(long int)
$!
$ OS
$ WS "int main(void)"
$ WS "{"
$ WS "  return sizeof(long int) != 8;"
$ WS "}"
$ CS

... other magic here ...
```
help ?

freddy77: > > help ?
> 
> I think FreeTDS should not attempt to define this, and let UnixODBC manage itself . I don't see what benefit is gained by trying to define it.

It's a bit more complicated. The issue is due to some unixODBC installations causing some very nasty issues. If unixODBC does not find `SIZEOF_LONG_INT` defined it falls back to include `unixodbc.h` header which contains configure time definition for some macros line `SIZEOF_LONG_INT`. The issue is that in some multiple-bit installations (like x86 and x64 architectures) there is a single `unixodbc.h` definition which is bound to the configure time setting. So you could end up with unixODBC using an ABI and FreeTDS (or other unixODBC client) using another. Results are pretty nasty (from data corruptions to buffer overflows).

Said that if OpenVMS does not have the multiple-bit issue you could avoid defining `SIZEOF_LONG_INT`.

include/odbcss.h
mmcnabb-vms: > @mmcnabb-vms just to confirm, are you find with the proposed `inline` change?

Yes, thanks

vms/config_h.vms
mmcnabb-vms: @freddy77 I'm dealing with that issue in the VMS port of UnixODBC by reworking `sqltypes.h` slightly so that there is no dependency on `sizeof(long)`. It's not a new issue to VMS users ever since VSI C++ was developed with sizeof(long) 8, whereas VSI C has sizeof(long) 4 -- all of the system headers had to be revised to avoid using `long`.    In order to support projects that use both C and C++, we literally can't have anything in headers that would depend on `sizeof(long)`  

Also, the way packages and installation locations work in VMS, it's easy to deploy database applications each with their own UnixODBC (and FreeTDS) builds, as opposed to having one system-wide installation .

PR #688, opened 2025-12-04
Minor changes for OpenVMS

freddy77:
* Remove need for file tweak, already fixed
   `%td` usage in format strings was removed by commit d01670c2ed63a91 (cfr `Do not use %td printf formatting string`);
* Simplify expression in C source tests
   Just pass the buffer, the type is already compatible, no need to cast.
* Use proper declaration for `main()` function in tests
    Do not list parameter list empty.

### DIFF
--- a/vms/configure.com
+++ b/vms/configure.com
@@ -226,7 +226,7 @@ $!
 $ OS
 $ WS "#include <stdio.h>"
 $ WS "#include <stdlib.h>"
-$ WS "int main()"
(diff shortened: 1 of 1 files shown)


freddy77:
@craigberry @mmcnabb-vms really minor changes just looking at the other PR

craigberry:
This all looks good and builds fine.

mmcnabb-vms:
Approve

PR #692, opened 2026-01-14
tds: fix potential NULL dereference

ProjectMutilation:
The result of the realloc function can be NULL if the memory allocation fails.
Found by PostgresPro with Svace Static Analyzer

### DIFF
--- a/src/tds/token.c
+++ b/src/tds/token.c
@@ -2541,7 +2541,7 @@ tds_alloc_get_string(TDSSOCKET * tds, char **string, size_t len)
 		*string = NULL;
 		return -1;
 	}
-	s = (char*) realloc(s, out_len + 1);
+	TDS_RESIZE(s, out_len + 1);
 	s[out_len] = '\0';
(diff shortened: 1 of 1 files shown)


ProjectMutilation:
> It cannot fail if the size is smaller than before.

A realloc behavior is implementation defined. The standard isn't warranty this.

freddy77:
This should do
```diff
diff --git a/src/tds/token.c b/src/tds/token.c
index 79f097ad1..c7826fb43 100644
--- a/src/tds/token.c
+++ b/src/tds/token.c
@@ -2541,7 +2541,7 @@ tds_alloc_get_string(TDSSOCKET * tds, char **string, size_t len)
                *string = NULL;
                return -1;
        }
-       s = (char*) realloc(s, out_len + 1);
+       TDS_RESIZE(s, out_len + 1);
        s[out_len] = '\0';
        *string = s;
        return 0;
```

freddy77 (approved):

Issue #686, opened 2025-11-17
Handling of NULL values in FreeBCP native file format with MSSQL


mmcnabb-vms:
### Overview 
`freebcp -n`  , and `freebcp -f formatfile.fmt` where the format file specifies the native type for the column, does not export and input NULL values correctly in some cases.  By "correctly" I mean:

- Rows should always roundtrip successfully, e.g. `out` on a table, clearing the table, then `in` from the same data file should exactly restore the table (subject to generation of default values where intended, of course, as per #685).

- A nice-to-have would be for the binary format of `freebcp -n` to match that of MSSQL or ASE `bcp -n`.
  - However this is not really important: the rationale for `-n` is for maximum performance locally, and (according to the documentation anyway) it's not even necessarily true that `bcp -n` on two different machines both talking to the same MSSQL instance will use the same format as each other.
  - The `-c` format is there to be used for "portable" data transfer.

### Observations from testing `VARCHAR` fields
Problems to be fixed are highlighted in bold:

- MSSQL `bcp -n` outputs NULL VARCHAR as `\xFF\xFF`
- MSSQL `bcp -n` outputs zero-length VARCHAR as `\x00\x00`
- ASE `bcp -n` outputs NULL VARCHAR as `\x00\x00`
- ASE does not support zero-length VARCHAR (any attempt to insert such a thing into the database actually inserts a length-1 string consisting of a single space) so there is no conflict of representations between NULL and empty-string, since the latter cannot exist in ASE.
- `freebcp -n` outputs `\x00\x00` for both NULL varchar, and empty strings in MSSQL, so it cannot reconstruct the original data correctly when importing. **When operating on MSSQL it should use the MSSQL format for NULL VARCHARs.**

### Observations from testing `DATETIME` fields

- MSSQL `bcp -n` outputs `\xFF` for a NULL DATETIME.  (Other values are `\x08` followed by 8 bytes;  the start-of-epoch date is represented by those 8 bytes all being zero).
- ASE `bcp -n` outputs `\x00` for a NULL DATETIME.
- `freebcp -n` outputs `\x00` for a NULL DATETIME. 
- `freebcp -n` in input mode, when the file contains `\x00` for a DATETIME field, causes corruption. The bcp operation may fail with EOF error or succeed; and if it succeeds, the number of rows can be smaller than it should be, and fields may have incorrect values (usually `0`, but sometimes apparently random large numbers and so on). **This should work properly.**
[...]

PR #695, opened 2026-01-27
Issue685 - support default values for BCP In on ASE 

mmcnabb-vms:
Summary of fixes on this PR:

- The first three commits fix bugs in bcp in/out , which are actually discussed on #686 . With these fixes in place, data should round-trip correctly using {freebcp|MSbcp|ASEbcp} with character format interchangeably on the same data file.  The roundtrip of `freebcp -n` with SQL Server has had some fixes but there are still outstanding issues regarding nulls in which are covered on 686 and I'll work on at a later date
- Added two new tests and slightly expanded `d_bcp` test relating to the handling of NULL and Default Values in bcp
- When talking to ASE: implement the feature of ASE bcp that if a column has a Default Value defined in the database, then the column can be omitted from the BCP format file, or have value specified as NULL in the data file where permitted, with the effect that the Default Value for the column will be inserted; even if the column was defined as non-nullable.    (MS bcp does not allow omitting columns, and handles default values differently in a way that's already supported by freebcp).
- Allow multiple Hints (that's the stuff in the `WITH()` clause). I also removed the code that validates the hints, since the database will reject an invalid hint and I don't see what we gain by rejecting the hint on client side; it seems more forwards-compatible to let the database handle this.
- Implement the `-k` flag as supported by MSBCP (ASE bcp calls it `--ignoreDefaults`) which means that even if a column has a default value defined, BCP inserting a NULL will actually insert NULL and not the Default Value.

Most of the content is covered by the new test cases, because `freebcp` is a fairly thin wrapper around DBlib bcp API. However I also have made a local test script that does a few "front door" tests calling `freebcp` directly to test the null/default handling, and did ad-hoc testing of `-k`.

### DIFF
--- a/src/apps/freebcp.c
+++ b/src/apps/freebcp.c
@@ -537,6 +537,7 @@ static int
 file_process(BCPPARAMDATA *pdata, DBPROCESS *dbproc, DBINT dir)
 {
 	DBINT li_rowsread = 0;
+	DBINT li_rowscopied = 0;
 	int i;
 	int li_numcols;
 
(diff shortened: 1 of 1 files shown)


mmcnabb-vms:
> It starts being time to prepare for a new release, there are quite some new features and improvements, especially in BCP code. I would like to integrate most of the BCP fixes/changes before the release.
> 
> Yes, a rebase on master would help.

OK. I see now Ucko PR #558 where he had already done the Defaults change, managed to overlook that when checking his other PRs.  The other changes on #558 look at first sight as if they could come in to master without too much trouble

PR #697, opened 2026-02-04
Issues found by static analysis

mmcnabb-vms:
Here are a bunch of minor issues identified by MSVC's static analyzer - please feel free to just pick out anything important, and/or squash multiple items in a batch. Not urgent and I don't mind if some are rejected.

The more significant ones:
- The first commit fixes a leak of a thread handle
- The second commit, I am not exactly sure if this might break some use cases, but the test suite still all passes.
- The tds_dir_snprintf I think would be a genuine overflow problem

Other notes
- There are quite a lot of build warnings on a Windows build (and some on a Linux build) regarding implicit narrowing integer conversions, and signed/unsigned mismatches. I would ideally like to fix all of these but maybe after some discussion about how exactly to go about it.
- I fixed issues raised by the analyzer on tds.lib and tdsodbc, but not on dblib or sybdb yet.

### DIFF
--- a/src/odbc/unittests/common.c
+++ b/src/odbc/unittests/common.c
@@ -29,6 +29,11 @@ struct odbc_buf{
 	void *buf;
 };
 
+/* This unit cannot be built with wide SQLTCHAR because it uses strcpy,
+ * printf etc. to handle the resulting strings. 
+ */
+TDS_COMPILE_CHECK(sqltchar_narrow, sizeof(SQLTCHAR) == 1);
+
 HENV odbc_env;
 HDBC odbc_conn;
 HSTMT odbc_stmt;
@@ -337,7 +342,7 @@ odbc_db_is_microsoft(void)
 
 	if (ms_db < 0) {
 		buf[0] = 0;
-		SQLGetInfo(odbc_conn, SQL_DBMS_NAME, buf, sizeof(buf), &len);
+		CHKGetInfo(SQL_DBMS_NAME, buf, sizeof(buf), &len, "S");
 		for (i = 0; buf[i]; ++i)
 			buf[i] = tolower(buf[i]);
 		ms_db = (strstr(C(buf), "microsoft") != NULL);
@@ -356,7 +361,7 @@ odbc_driver_is_freetds(void)
 
 	if (freetds_driver < 0) {
 		buf[0] = 0;
-		SQLGetInfo(odbc_conn, SQL_DRIVER_NAME, buf, sizeof(buf), &len);
+		CHKGetInfo(SQL_DRIVER_NAME, buf, sizeof(buf), &len, "S");
 		for (i = 0; buf[i]; ++i)
 			buf[i] = tolower(buf[i]);
 		freetds_driver = (strstr(C(buf), "tds") != NULL);
@@ -455,7 +460,7 @@ odbc_tds_version(void)
 	if (odbc_driver_is_freetds() && tds_version < 0) {
 		version = 0;
 		len = 0;
-		SQLGetInfo(odbc_conn, 1300 /* SQL_INFO_FREETDS_TDS_VERSION */, &version, sizeof(version), &len);
+		CHKGetInfo(1300 /* SQL_INFO_FREETDS_TDS_VERSION */, &version, sizeof(version), &len, "S");
 		if (len == sizeof(version))
 			tds_version = (version >> 16) << 8 | (version & 0xff);
 	}


include/freetds/macros.h
```diff
@@ -64,6 +64,11 @@
 #define TDS_WUR __attribute__ ((__warn_unused_result__))
 #else
 #define TDS_WUR
+#endif
+
+ /* Microsoft Static Analyzer Language (windows.h defines _Field_size_) */
+#ifndef _Field_size_
+#define _Field_size_(x)
```
freddy77: I would avoid using reserved identifiers on different systems

src/odbc/odbc_util.c
```diff
@@ -434,7 +434,7 @@ odbc_set_string_flag(TDS_DBC *dbc TDS_UNUSED, SQLPOINTER buffer, SQLINTEGER cbBu
 			result = SQL_ERROR;
 		out_len = cbBuffer - (int) ol;
 		while (result != SQL_ERROR && il) {
-			char discard[128];
+			char discard[128] = { 0 };	/* tds_iconv sometimes does not write output, e.g. if il==0 */
```
freddy77: Please not, just a waste of CPU

src/tds/mem.c
```diff
@@ -1098,7 +1098,10 @@ TDSPACKET *
 tds_realloc_packet(TDSPACKET *packet, unsigned len)
 {
 	if (packet->capacity < len) {
-		packet = (TDSPACKET *) realloc(packet, len + TDS_OFFSET(TDSPACKET, buf));
+		void* newpacket = realloc(packet, len + TDS_OFFSET(TDSPACKET, buf));
+		if (!newpacket)
+			free(packet);
+		packet = newpacket;
```
freddy77: This is just a false positive, `packet` is passed, what is changed here is a local variable, the called still have the valid old pointer.

src/tds/net.c
```diff
@@ -1141,8 +1142,8 @@ tds7_get_instance_ports(FILE *output, struct addrinfo *addr)
 	 */
 	for (num_try = 0; num_try < 16 && msg_len == 0; ++num_try) {
 		/* send the request */
-		msg[0] = 3;
-		if (sendto(s, msg, 1, 0, addr->ai_addr, addr->ai_addrlen) < 0)
+		char msg_type = 3;
+		if (sendto(s, &msg_type, 1, 0, addr->ai_addr, addr->ai_addrlen) < 0)
```
freddy77: What is the complaining here ? Maybe `&msg[0]` ?

src/tds/packet.c
```diff
@@ -429,7 +430,9 @@ tds_connection_network(TDSCONNECTION *conn, TDSSOCKET *tds, int send)
 static TDSRET
 tds_connection_put_packet(TDSSOCKET *tds, TDSPACKET *packet)
 {
-	TDSCONNECTION *conn = tds->conn;
+	TDSCONNECTION* conn;
+	assert(tds != NULL);
+	conn = tds->conn;
```
freddy77: I don't see the reason

src/odbc/unittests/common.c
```diff
@@ -29,6 +29,11 @@ struct odbc_buf{
 	void *buf;
 };
 
+/* This unit cannot be built with wide SQLTCHAR because it uses strcpy,
+ * printf etc. to handle the resulting strings. 
+ */
+TDS_COMPILE_CHECK(sqltchar_narrow, sizeof(SQLTCHAR) == 1);
```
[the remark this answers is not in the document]
mmcnabb-vms: The test driver builds with the MSVC character set option of MBCS (not UNICODE), whether or not you have defined `ENABLE_ODBC_WIDE`.  And it's a non-issue in Linux since everyone uses narrow API and UTF-8. 

Maybe it is a gap in test coverage if we are not actually testing the ODBC wide versions anywhere ... 

Updating the test driver to allow SQLWCHAR will be a bit of a pain, since it does things like copy SQLSTATE into a global `char[6]` , which won't work if the return SQLSTATE is 6 wide characters.  A lot of the test driver code would need to have MBCS <-> wchar conversion functions added.

src/tds/mem.c
mmcnabb-vms: At two of the call sites, the calling code is `packet = tds_realloc_packet(packet, ....);`  , so if `realloc` returns NULL then memory is leaked in those cases. Not a huge deal though as if realloc is failing we have bigger issues. 

So we could leave `tds_realloc_packet` unchanged, and optionally fix those call sites.

include/freetds/macros.h
```diff
@@ -82,4 +87,15 @@
 #define tds_new(type, n) ((type *) malloc(sizeof(type) * (n)))
 #define tds_new0(type, n) ((type *) calloc(n, sizeof(type)))
 
+/* Tell the compiler that a function never returns. Use before the prototype */
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
+#define TDS_NORETURN _Noreturn
```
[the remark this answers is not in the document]
mmcnabb-vms: It was only added in C11 and not everyone is using C11 compilers.  Another option might be to #define _Noreturn on pre-C11 systems , that's not really substantially different to what I've put though.

src/tds/mem.c
freddy77: I think then fixing the caller is better, something like
```diff
diff --git a/src/tds/packet.c b/src/tds/packet.c
index ec6840991..4bfa2ce02 100644
--- a/src/tds/packet.c
+++ b/src/tds/packet.c
@@ -259,7 +259,7 @@ tds_packet_read(TDSCONNECTION *conn, TDSSOCKET *tds)
 Memory_Error:
 Severe_Error:
        tds_connection_close(conn);
-       tds_free_packets(packet);
+       tds_free_packets(conn->recv_packet);
        conn->recv_packet = NULL;
        return false;
 }
```

src/tds/packet.c
freddy77: Yes, but `tds_connection_put_packet` is called from 2 places and in both `tds` is not `NULL`.

src/utils/tds_cond.c
```diff
@@ -114,9 +114,12 @@ detect_cond(void)
...
 
-	init_cv  = (init_cv_t)  GetProcAddress(mod, "InitializeConditionVariable");
-	sleep_cv = (sleep_cv_t) GetProcAddress(mod, "SleepConditionVariableCS");
-	wake_cv  = (wake_cv_t)  GetProcAddress(mod, "WakeConditionVariable");
+	if (mod)
+	{
+		init_cv = (init_cv_t)GetProcAddress(mod, "InitializeConditionVariable");
+		sleep_cv = (sleep_cv_t)GetProcAddress(mod, "SleepConditionVariableCS");
+		wake_cv = (wake_cv_t)GetProcAddress(mod, "WakeConditionVariable");
+	}
```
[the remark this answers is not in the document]
freddy77: Merged this part

include/freetds/macros.h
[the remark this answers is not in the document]
freddy77: applied

src/odbc/odbc.c
```diff
@@ -1742,7 +1743,9 @@ odbc_SQLAllocConnect(SQLHENV henv, SQLHDBC FAR * phdbc)
 	tds_dstr_init(&dbc->attr.current_catalog);
 	dbc->attr.login_timeout = 0;	/* TODO */
 	dbc->attr.metadata_id = SQL_FALSE;
+	/* This is deprecated and Microsoft recommends allowing the default to apply.
```
freddy77: It does not make much sense, `SQL_CUR_USE_IF_NEEDED` value is zero so this change is a no-op.
Why not changing the default to `SQL_CUR_USE_DRIVER` ? We support server cursors (at least for MSSQL).

src/tds/config.c
```diff
@@ -609,6 +609,9 @@ tds_parse_conf_section(const char *option, const char *value, void *param)
 
 	tdsdump_log(TDS_DBG_INFO1, "\t%s = '%s'\n", option, value);
 
+	if (!param)
+		return false;
```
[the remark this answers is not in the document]
freddy77: The ODBC connection string is more similar to MS attributes. There's a `TrustServerCertificate` and a `HostNameInCertificate`. Anyway, if you feel that ignoring just the name has value we can add an additional attribute.

src/odbc/odbc_util.c
freddy77: Given the weirdness of `iconv` prototype (not common to have bytes left passed by reference), yes, it's a false positive.

src/tds/packet.c
freddy77: In Unix usually the `assert` is compiled in even in production. I think many compilers have an attribute nowadays to declare that the pointer should not be NULL. Isn't there something equivalent in MSVC ?

src/tds/config.c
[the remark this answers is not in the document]
freddy77: @mmcnabb-vms about the original commit, now the compiler should not complain. Can you test it?

src/tds/mem.c
mmcnabb-vms: Yes, that was my understanding of this interaction (instead of my change, do your call site fix)

src/tds/packet.c
freddy77: There's a `tds_extra_assert` which should never be used in production. Not exactly the same. Not much worried about "extreme" NULL pointer dereference. Program crash cleanly.
I saw there are `_In_opt_` and similar indicating a possible NULL pointer or not.
It looks kind of ugly adding all these notes anyway (and surely not using reserved identifiers).

src/odbc/odbc.c
mmcnabb-vms: Has been merged by Freddy

PR #703, opened 2026-02-17
BCP defaults

mmcnabb-vms:
Changes related to Sybase/ASE BCP and BCP Defaults

- Some bugfixes to my previous submissions about LOCK DATAROWS
- Support "-k" flag meaning to not apply defaults even if they exist
- Test the skipping of columns in BCP. This is a Sybase feature - a format file can be used to do a "bcp in" with the data file only containing some columns; and remaining columns should behave as if NULL were being inserted, including applying default if default exists.


mmcnabb-vms:
Rebased to master.  One line of code had to be changed as `convert.h` was moved.

Note, the first four commits are not to do with BCP defaults; they are fixes to ASE datarows-locking support, including fixing a buffer overflow.  Could be merged first before considering the rest of this PR.

src/tds/bulk.c
```diff
@@ -199,12 +199,13 @@ tds_bcp_init(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
...
-		if (!bindinfo->current_row)
-			goto cleanup;
-		bindinfo->row_free = tds_bcp_row_free;
-	}
+	/* bindinfo->current_row is used to hold TDS5 BCP row packing buffer, we will
+	 * allocate some space later once we know how big the row will be. 
+	 * Not to be confused with the same members of "resinfo"
+	 * in a normal (non-bcp) query,  which holds unpacked row data.
+	 */
+	bindinfo->current_row = NULL;
```
freddy77: This look weird written like this. If the pointer was not `NULL` this looks like a leak, if it was `NULL` it looks like a no-operation. I would just remove it, the pointer is `NULL`.

src/tds/bulk.c
```diff
@@ -199,12 +199,13 @@ tds_bcp_init(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
...
+	bindinfo->row_size = 0;
```
freddy77: I would just remove the initialization line above.

src/tds/bulk.c
```diff
@@ -468,6 +479,7 @@ tds_bcp_start_insert_stmt(TDSSOCKET * tds, TDSBCPINFO * bcpinfo)
 		if (erc < 0)
 			return TDS_FAIL;
 	} else {
+		/* NOTE: Current ASE docs do not mention "insert bulk", it might be deprecated? */
```
freddy77: `insert bulk` was not even documented for MSSQL, I think it was just considered an "internal" query. So I won't say it's _deprecated_ but just _undocumented_.

src/tds/log.c
```diff
@@ -488,49 +489,72 @@ tdsdump_log_impl(const char* file, unsigned int level_line, const char *fmt, ...
...
-	char* data;
+	char buf[30] = "NULL";
+	char* data = buf;
+	int data_len = -1;
 	TDS_SMALLINT type;
-	
+	TDSDATEREC dr;
+	TDS_NUMERIC* pnumeric;
+	TDS_MONEY* pmoney;
+	TDS_MONEY4* pmoney4;
```
freddy77: Won't it be easier to just use `tds_convert` ? You can use `TDS_CONVERT_CHAR` to avoid allocation and to write directly in `buf`.
This function is used only for logging, not expected to be optimized for speed.

src/apps/freebcp.h
```diff
@@ -68,6 +68,7 @@ typedef struct pd
 	int Tflag;
 	int Aflag;
 	int Eflag;
+	int ignoreDefaults;	/* Really insert a NULL even if default value defined */
```
freddy77, line 71: `bool` ?

src/tds/bulk.c
```diff
@@ -1179,7 +1201,10 @@ tds5_read_bulk_defaults(TDSRESULTINFO *res_info, TDSBCPINFO *bcpinfo)
 				break;
 			++syb_info;
 		}
-
+#if ENABLE_EXTRA_CHECKS
```
freddy77, line 1196: This code is not a check

src/tds/mem.c
```diff
@@ -1831,7 +1831,13 @@ tds_alloc_bcp_column_data(unsigned int column_size)
 {
 	BCPCOLDATA *coldata;
 
-	TEST_MALLOC(coldata, BCPCOLDATA);
+    /* Initialize fields to NULL, otherwise any field with value not
+     * explicitly set will behave as a non-null value of length 0,
+     * which is not really what we ever want.
+     * ASE BCP treats uninitialized fields as null.
+     */
+	TEST_CALLOC(coldata, BCPCOLDATA, 1);
```
freddy77, line 1839: `TEST_MALLOC` already use `calloc`.

src/tds/mem.c
```diff
@@ -1831,7 +1831,13 @@ tds_alloc_bcp_column_data(unsigned int column_size)
...
+    coldata->is_null = true;
```
freddy77, line 1840: Indeed, it's a good idea setting `NULL` as default.

freddy77:
Merged first 2 commits (minor style changes).

freddy77:
Merged most of the PR. Just 4 smaller commit left (see open conversations above).

src/tds/bulk.c
mmcnabb-vms: OK. `bindinfo` was allocated by `TEST_MALLOC` which resolves to `calloc` (must be some history there...) My thought with this code was to just clearly show the initialization of `current_row` and `row_size`. No big deal

src/tds/bulk.c
mmcnabb-vms: Well, OK.  I don't think this code should be enabled in a default build both for security reasons (showing the content of all fields) as well as bloating the log. Perhaps there is some other flag it could be behind?

src/tds/log.c
mmcnabb-vms: No objection to changing this up if there's an easier way.  It's a few months ago now but I think I didn't see a way to use `tds_convert` here at the time.

mmcnabb-vms:
> Merged most of the PR. Just 4 smaller commit left (see open conversations above).

Cool. I'm away the next two weeks,  go ahead as you see fit on the remaining commits if you like.

src/tds/bulk.c
mmcnabb-vms: OK. In that case perhaps my additional comment lines 202-205 would be better suited to a commit message than a code comment then

src/tds/bulk.c
mmcnabb-vms: OK; suggest wording the code comment to note it as an undocumented feature

mmcnabb-vms:
Re. the four outstanding commits:

- 406fc3b, 9a0d716, 24afcc7 I have submitted alternative commits along the lines your comments suggested (I decided to push a Revert and a new fix, instead of squashing, so that conversation doesn't get orphaned -- appveyor doesn't like it but the new commits cherry-pick to master with no issues for me)
- f83a8df : awaiting your thoughs on that one, it's just a question of what to call the guard that tdsdump_col is behind. It could just be #if 0 or otherwise commented out .

src/tds/log.c
mmcnabb-vms: The reason I avoided it at first was that `tds_convert` requires a `TDSCONTEXT` , but the existing call sites of `tdsdump_col` did not supply a context

freddy77: I have a change, I still need to properly test it.

src/tds/bulk.c
freddy77: Could be
```c
		/* NOTE: Current ASE docs do not mention "insert bulk", probably an undocumented internal syntax, as MS used to have it */
```

src/tds/log.c
mmcnabb-vms: OK. I coded your tds_convert suggestion but just go with your preference

freddy77:
>     * [f83a8df](https://github.com/FreeTDS/freetds/commit/f83a8dfac5fa26b9d5bff07cf6768229063a859b) : awaiting your thoughs on that one, it's just a question of what to call the guard that tdsdump_col is behind. It could just be #if 0 or otherwise commented out .

At the moment I used a new `TDS_DEBUG_DATA` that you need to set manually.
Maybe worth adding an option in `configure`/`cmake` (kind of `--enable-insecure-dump`).

src/tds/bulk.c
freddy77: I would like to close this pull request.
I suppose you can post another one about this.

src/tds/bulk.c
```diff
@@ -199,13 +199,6 @@ tds_bcp_init(TDSSOCKET *tds, TDSBCPINFO *bcpinfo)
 			goto cleanup;
 	}
 
-	if (!IS_TDS7_PLUS(tds->conn)) {
-		bindinfo->current_row = tds_new(unsigned char, bindinfo->row_size);
-		if (!bindinfo->current_row)
-			goto cleanup;
-		bindinfo->row_free = tds_bcp_row_free;
-	}
-
```
freddy77, line 208: This change was causing a straight forward crash. Fixed in the commit.

freddy77:
If there are no objections I would close this PR.

mmcnabb-vms:
All changes merged

PR #719, opened 2026-03-31
Make greater use of GNUInstallDirs

Zopolis4:
`GNUInstallDirs` has been notionally in use since 2010478fb247fdcae878bac2ad4cf10c44a5dd07, but it was never applied to the vast majority of install targets.


freddy77:
Merged, thanks

PR #721, opened 2026-04-07
ODBC date/time format strings

### LINKED ISSUE #715: ODBC date-to-char conversion
This issue regards the case of using `SQLBindCol` or `SQLGetData` where the database column is a date/time, but the bound data type is a character array. 

The conversion is handled inside `src/odbc/odbc.c` function `odbc_tds2sql()` where there is the code: `fmt = "%Y-%m-%d %H:%M:%S.%z"` .   As far as I can tell, the ODBC specification does not specify any particular format string here, it is just up to the driver's discretion.

We would like to make this datetime format configurable, as it assists people migrating from other databases where the client uses a different format.  

It's straightforward to code, but my question is what is the best way to specify this new configuration parameter in the FreeTDS configuration? 

Given that the FreeTDS Locales documentation suggests `locales.conf` is being phased out in favour of `freetds.conf` options, I am thinking of just adding new configuration parameters called `odbc date format` , `odbc date-only format`, `odbc time-only format` for example. 

Another option might be a section `[odbc]` that is parsed using the current locale parser; as that does have options for date, time, and datetime conversion strings, but that might be inflexible going forward, if it turns out that other ODBC conversion tweaks are desired; and also wouldn't be able to be configured differently for different data sources.

Another way could be custom ODBC configuration parameters along the same lines, instead of custom freetds.conf parameters, although it seems a bit weird to have such a thing in a connection string (for users that use connection strings without ODBC DSNs) and no other driver except FreeTDS would recognize the custom parameters anyway.

mmcnabb-vms:
Allow specifying the format string for ODBC implicit datetime-to-char conversions. Fixes #715.  The use case is for people migrating to TDSODBC from other database clients with different date/time formatting.

Configuration Examples :
- C API: `SQLSetConnectAttr(hdbc, SQL_COPT_TDSODBC_DATETIME_FORMAT, "%Y%m%d%H%M%S", SQL_NTS);`
- Defining a connection string `DRIVER=FreeTDS;......;DATETIMEFMT={%Y%m%d%H%M%S}`

Not urgent to review and merge this, so long as you're happy with the form of the above configuration methods.

The changes:
* Add new connection attribute to set date, time, or datetime format string. This setting is saved in the DBC.
* Add connection string attributes to match.
* Use these format strings instead of hardcoded strings in `odbc_tds2sql()`.
* Update test suite to include coverage of the new features.
* Redo the code to compute SQL_DESC_LENGTH and SQL_DESC_DISPLAY_SIZE to return a value based on the current format string.  Also, this implicitly fixes what appears to be a bug in the MS ODBC driver regarding the length of smalldatetime. 

No matching attribute has been added for ODBC DSNs or for freetds.conf. Does not affect conversion of char to datetime.


mmcnabb-vms:
Repushed - Linux build has a script that checks connection string parameter names are in alphabetical order in `include/freetds/odbc.h`

mmcnabb-vms:
In TDS 7.2 and older, date/time/datetime2 do not exist as wire formats (NVARCHAR is used instead), but `datetime` does exist. This formatting change only applies to conversion of datetime wire formats, as we have no way of knowing if an arriving NVARCHAR was originally a date or not.

Updated o_date test to check TDS version before testing those formats. 

A problem is that in TDS 7.2 and older, SQL_DESC_LENGTH would be incorrect for `datetime2`, but correct for `datetime`, if the user specifies a datetime replacement format . I don't think we have any way to address this in FreeTDS since the source type is determined by the query string and/or database structure; the server converts the date to NVARCHAR before it gets to us.  It's something user would have to be aware of if they are using TDS 7.2 or older.

mmcnabb-vms:
Rebased to master - No conflicts

src/odbc/connectparams.c
```diff
@@ -492,6 +492,9 @@ odbc_parse_connect_string(TDS_ERRS *errs, const char *connect_string, const char
 			   ||  strcasecmp(option, "client_charset") == 0) {
 			num_param = ODBC_PARAM_ClientCharset;
 			tds_parse_conf_section(TDS_STR_CLCHARSET, tds_dstr_cstr(&value), login);
+		} else if (CHK_PARAM(DateTimeFmt)) {
```
freddy77, line 495: It's a bit weird, all other assign the field directly. Maybe pass `dbc` to this function and use it if not `NULL` ?

More high level. Should these attributes be per connection or per environment ?

src/odbc/odbc.c
```diff
@@ -572,6 +572,19 @@ odbc_prepare(TDS_STMT *stmt)
...
+/** Helper for SQLDriverConnect to apply DBC parameters */
+static void dbc_apply_param(TDS_DBC* dbc, DSTR* dst, const TDS_PARSED_PARAM* param)
+{
+	if (param->len > 0)
+	{
+		/* TDS_PARSED_PARAM still contains the optional {} delimiters */
+		int offset = (param->p[0] == '{' && param->p[param->len - 1] == '}');
+
+		if (!tds_dstr_copyn(dst, param->p + offset, param->len - 2 * offset))
+			odbc_errs_add(&dbc->errs, "HY001", NULL);
```
freddy77, line 584: This looks caused by not saving these parameters in `odbc_parse_connect_string`. Also this is not removing the quote correctly. Think about something like `{hello}}world}`.

src/odbc/odbc.c
```diff
@@ -1766,6 +1783,15 @@ odbc_SQLAllocConnect(SQLHENV henv, SQLHDBC FAR * phdbc)
...
 	dbc->attr.bulk_enabled = SQL_BCP_OFF;
 
+	/* Initial values of DBC date formats taken from environment context.
+	 * odbc_SQLAllocEnv() always initializes that. */
+	tds_dstr_init(&dbc->datetime_fmt);
+	tds_dstr_copy(&dbc->datetime_fmt, env->tds_ctx->locale->datetime_fmt);
+	tds_dstr_init(&dbc->date_fmt);
+	tds_dstr_copy(&dbc->date_fmt, env->tds_ctx->locale->date_fmt);
+	tds_dstr_init(&dbc->time_fmt);
+	tds_dstr_copy(&dbc->time_fmt, env->tds_ctx->locale->time_fmt);
```
freddy77, line 1793: Adding out of memory checks.

src/odbc/odbc.c
freddy77: Changed

src/odbc/odbc.c
freddy77: Added

src/odbc/connectparams.c
freddy77: Done (first sentence).

freddy77:
Merged with changes mentioned in comments and lot of minor style updates.

PR #729, opened 2026-06-14
Support Sybase syntax for defncopy, enabling VMS-style wrapper to work

craigberry:
FreeTDS never supported the documented Sybase syntax for the defncopy command.  Now it can via detecting the presence of an `in` or `out` parameter.  This may be a plus in its own right, but was necessary to get the VMS-style command wrapper working.

### DIFF
--- a/src/apps/defncopy.c
+++ b/src/apps/defncopy.c
@@ -916,32 +916,35 @@ print_results(DBPROCESS *dbproc)
 static void
 usage(const char invoked_as[])
 {
+
 	fprintf(stderr, "usage:  %s \n"
-			"        [-U username] [-P password]\n"
-			"        [-S servername] [-D database]\n"
-			"        [-i input filename] [-o output filename]\n"
-			"        [owner.]object_name [[owner.]object_name...]\n"
-			, invoked_as);
-/**
-defncopy Syntax Error
-Usage: defncopy
-    [-v]
-    [-X]
---  [-a <display_charset>]
---  [-I <interfaces_file>]
---  [-J [<client_charset>]]
---  [-K <keytab_file>]
-    [-P <password>]
---  [-R <remote_server_principal>]
-    [-S [<server_name>]]
-    [-U <user_name>]
---  [-V <security_options>]
---  [-Z <security_mechanism>]
---  [-z <language>]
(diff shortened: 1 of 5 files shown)


craigberry:
The build that fails looks like it's with the CI tools, not anything I changed:

`Compatibility with CMake < 3.5 has been removed from CMake.`

mmcnabb-vms:
Matching Sybase syntax seems eminently sensible to me.

In fact I'd propose keeping it simple by not having an environment variable, and just making the Sybase syntax be the syntax.   

It seems like it would be straightforward to support both syntaxes, with the old syntax being considered deprecated: e.g. if, `-i` or `-o` are present then stop looking for `in` and `out`.

craigberry:
> Matching Sybase syntax seems eminently sensible to me.
> 
> In fact I'd propose keeping it simple by not having an environment variable, and just making the Sybase syntax be the syntax.
> 
> It seems like it would be straightforward to support both syntaxes, with the old syntax being considered deprecated: e.g. if, `-i` or `-o` are present then stop looking for `in` and `out`.

I started of doing that but realized I couldn't. The trick is that `-i` and `-o` are not required and if absent default to stdin and stdout.  The existing FreeTDS implementation was designed to be a filter with nothing required but one or more object names.  Trying not to break that use case is why I put in the environment variable to switch syntax types.

freddy77:
Why the syntax cannot automatically detected using first no-option argument (only `in` or `out`) ?

craigberry:
> Why the syntax cannot automatically detected using first no-option argument (only `in` or `out`) ?

I'll look into it.  If anyone using the old syntax had database objects named `in` or `out` they'd be in trouble, but that's a pretty narrow corner case.  We'll have to print usage for both ways.

freddy77:
I suppose one could force one or another quoting the object name, like `.in` or `[in]`.

craigberry:
> I suppose one could force one or another quoting the object name, like `.in` or `[in]`.

Yes, quoting is a good workaround in the unlikely even anyone stumbles over it.  I have force pushed an update that removes the necessity of an environment variable.

freddy77:
Merged with minor space updates

PR #731, opened 2026-07-05
BCP input with format-file 

mmcnabb-vms:
For BCP input with format files:

- support this in test suite and add a test case
- perform more validation that the format file doesn't specify contradictory requirements
- validate length prefix, if present, on fixed fields
- reject terminated fields also having a length prefix, other than a null indicator  (unclear what this means anyway, considering that iconv may be happening)
- improve code commentary
- Add dry-run option for "bcp in" to check hostfile for fatal errors without doing a partial database write in case there were errors.

Semantic changes:
- when reading a variable field from the hostfile, that has a prefix or terminator, the prefix/terminator governs how many bytes to read from the hostfile. It should not perform a short read based on format file's column size, nor server column size.  It will truncate the data after reading, to match the server column size. (Matches behaviour of MS BCP)
- format file specifying column size 0 is only interpreted as null now for fixed fields with no prefix/terminator (Matches behaviour of MS BCP). This field is disregarded when the size is known in another way (fixed type, or variable type with a prefix or terminator).

Regarding iconv:  the existing code only performed iconv for a character field with a terminator; otherwise data was read raw. This commit does not change that behaviour, but documents it more clearly in the code.  

The previous code had comments indicating an intent to perform iconv for all cases of character fields; but (a) that's not so easy to implement as our stream read function does not have the option to read a set number of bytes and perform iconv , and (b) I don't think we actually want to do that anyway; because if a field does not have a terminator it means it's probably part of either a native-mode load, or a format file that behaves like native-mode, which means we want maximum performance by doing raw data in/out without conversion.

I did some ad-hoc testing of cases that were expected to cause failure,  since the test suite currently doesn't provide for a case that is expected to fail.

### DIFF
--- a/src/dblib/bcp.c
+++ b/src/dblib/bcp.c
@@ -1100,6 +1100,7 @@ _bcp_convert_in(DBPROCESS *dbproc, TDS_SERVER_TYPE srctype, const TDS_CHAR *src,
 	coldata->datalen = len;
 	if (variable) {
 		if (len > col->on_server.column_size) {
+			free(cr.c);
 			dbperror(dbproc, SYBECOFL, 0);
 			return TDS_FAIL;
 		}
--- a/src/dblib/bcp.c
+++ b/src/dblib/bcp.c
@@ -556,6 +556,9 @@ bcp_control(DBPROCESS * dbproc, int field, DBINT value)
 	case BCPBATCH:
 		dbproc->hostfileinfo->batch = value;
 		break;
+	case BCPDRYRUN:
+		dbproc->bcpinfo->dry_run = value;
+		break;
 
 	default:
 		dbperror(dbproc, SYBEIFNB, 0);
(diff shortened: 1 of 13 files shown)


mmcnabb-vms:
Failing build is a problem with CMake version mismatch in the CI pipeline .  I am using CMake 4.3.1-msvc1 in Windows with no problem. (The version that is deployed with Visual Studio)

src/dblib/bcp.c
```diff
@@ -1391,6 +1442,10 @@ _bcp_read_hostfile(DBPROCESS *dbproc, TDSFILESTREAM *stream, bool *row_error, bo
 		}
 		free(coldata);
 	}
+
+	if (dbproc->bcpinfo->dry_run && truncation_warning)
+		printf("Warning: data truncated due to server column being too small; see log file for detail.\n");
```
freddy77, line 1447: The library should not use `printf`.

freddy77:
Merged most but last 4 seems to introduce a regression, bcp test started failing.

freddy77:
Merged

freddy77:
> Merged most but last 4 seems to introduce a regression, bcp test started failing.

At the end turned out to be just noise.

